PR #56, opened 2015-05-07
Added flatten

JacobEdelman:


Hardmath123:
(See comments on IRC about how this isn't actually more efficient!)

Issue #58, opened 2015-05-08
Named tokens


Hardmath123:
Provide a way to name a token, and bind these names by augmenting the data array to an object.

```
a -> name:string {% function(d) {return d.name;} %}
```

matthew-dean:
So... Okay, so I think I was reading this backwards before (and hence my backwards examples). You're saying ```string``` is the token name (and not a type), and name is the object label assigned to the token?

In other words:
```
a -> space:_ {% function(d) {return d.space;} %}
# returns " "
```
?

Okay, I think I'm getting it now. So, in a more complex (but not ideal grammar) example:
```
a -> id:[\w]:+ _ count:[0-9]:*

# "cow 999" returns { id: ['c','o','w'], count: ['9','9','9'] }
```
Correct?

Hardmath123:
Yes, precisely, that's how I meant it originally. However, rather than returning the object `{ id: ['c','o','w'], count: ['9','9','9'] }`, it would actually return the array with the relevant properties set (because you can do gunk like that in JavaScript).

```js
result.constructor => Array
result.id => 'cow'
result.count => 999
result => ['cow', 999]
```

By the way, I think that's a pretty general issue with using colons—it's not immediately obvious whether the left side is the label or the nonterminal name. What did you think of the dot syntax?

```
a -> astringnonterminal.name _ anintegernonterminal.count {% function(d) {return [d.name, d.count]; } %}
```

matthew-dean:
Hmm, it's still kind of weird-looking to me, because there are so many template languages that use that to explicitly refer to the property of an object. So, I still had to think about what was the label and what was the nonterminal, even though it was explicit. That's why I think separating them entirely from grammar is smarter, because it's actually more explicit, and it keeps the grammar itself more readable. This, to me, is really easy to follow.

```
a -> astringnonterminal _ anintegernonterminal  {name, null, count}  {% function(d) {return [d.name, d.count]; } %}
```
I can read the grammar as grammar, the object / array assignment as one piece, and the processing function as the next. They're not semantically confused with each other. Right now, the grammar is JUST about parsing rules. Sticking labeling inline, I don't know. It just seems like it mucks it up and makes the grammar itself more confusing, and still leaves the key assignment still somewhat confusing, so I don't think putting it inline is a win.

I think you could also use the assignment form to further clean up long grammar in a more "documentation" style.
```
a -> astringnonterminal      {name}
     _                       {}
[...]

PR #63, opened 2015-07-16
postprocessors can have context

jlarsson:
I short, its possible to pass an optional context to Parser:

```
// js
var context = {foo: function (d) { ... }}
var p = new nearley.Parser(grammar.ParserRules, grammar.ParserStart, context)

...
// grammar
rule -> a b c {% function (d) { return this.foo(d) } %}
```

The motivation is that it in general makes it easier to construct more semantically bound ASTs and in particular allows complex logic to separated from the grammar specification.

I made some changes to the test structure:
- existing tests are now pre tests
- added unittests using mocha

Thanks for a great piece of work ;)

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -156,9 +156,9 @@ State.prototype.process = function(location, table, rules, addedRules) {
 
 
 
-function Parser(rules, start) {
+function Parser(rules, start, context) {
     var table = this.table = [];
-    this.rules = rules.map(function (r) { return (new Rule(r.name, r.symbols, r.postprocess)); });
+    this.rules = rules.map(function (r) { return (new Rule(r.name, r.symbols, r.postprocess ? r.postprocess.bind(context) : r.postprocess)); });
     this.start = start = start || this.rules[0].name;
     // Setup a table
     var addedRules = [];
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "compiler"
   ],
   "scripts": {
-    "test": "npm install && sh build.sh"
+    "pretest": "npm install && sh build.sh",
+    "test": "sh test/unittests/build.sh && mocha test/unittests/**.js"
   },
   "author": "Hardmath123",
   "contributors": [
@@ -30,5 +31,8 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/hardmath123/nearley.git"
+  },
+  "devDependencies": {
+    "mocha": "^2.2.5"
   }
 }
--- a/test/unittests/build.sh
+++ b/test/unittests/build.sh
@@ -1,0 +1,1 @@
+bin/nearleyc.js test/unittests/context-grammar.ne -o test/unittests/context-grammar.js;
--- a/test/unittests/context-grammar.js
+++ b/test/unittests/context-grammar.js
@@ -1,0 +1,32 @@
+// Generated automatically by nearley
+// http://github.com/Hardmath123/nearley
+(function () {
+function id(x) {return x[0]; }
+var grammar = {
+    ParserRules: [
+    {"name": "program", "symbols": ["_", "uc", "_"], "postprocess":  function (d) { return this.id(d[1]) } },
+    {"name": "program", "symbols": ["_", "lc", "_"], "postprocess":  function (d) { return this.id(d[1]) } },
+    {"name": " string$1", "symbols": [{"literal":"t"}, {"literal":"o"}, {"literal":"u"}, {"literal":"p"}, {"literal":"p"}, {"literal":"e"}, {"literal":"r"}], "postprocess": function joiner(d) {
+        return d.join('');
+    }},
+    {"name": "uc", "symbols": [" string$1", "_", "string"], "postprocess":  function (d) { return this.toUpper(d[2]) } },
+    {"name": " string$2", "symbols": [{"literal":"t"}, {"literal":"o"}, {"literal":"l"}, {"literal":"o"}, {"literal":"w"}, {"literal":"e"}, {"literal":"r"}], "postprocess": function joiner(d) {
+        return d.join('');
+    }},
+    {"name": "lc", "symbols": [" string$2", "_", "string"], "postprocess":  function (d) { return this.toLower(d[2]) } },
+    {"name": "string", "symbols": [{"literal":"'"}, " ebnf$3", {"literal":"'"}], "postprocess":  function (d) { return this.string(d[1]) } },
+    {"name": "_", "symbols": [], "postprocess":  function(d) { return this.null(); } },
+    {"name": "_", "symbols": [/[\s]/, "_"], "postprocess":  function(d) {return this.null(); } },
+    {"name": " ebnf$3", "symbols": []},
+    {"name": " ebnf$3", "symbols": [/[a-zA-Z ]/, " ebnf$3"], "postprocess": function (d) {
+                    return [d[0]].concat(d[1]);
+                }}
+]
+  , ParserStart: "program"
+}
+if (typeof module !== 'undefined'&& typeof module.exports !== 'undefined') {
+   module.exports = grammar;
+} else {
+   window.grammar = grammar;
+}
+})();
--- a/test/unittests/context-grammar.ne
+++ b/test/unittests/context-grammar.ne
@@ -1,0 +1,11 @@
+program -> _ uc _ {% function (d) { return this.id(d[1]) } %}
+        | _ lc _ {% function (d) { return this.id(d[1]) } %}
+
+uc -> "toupper" _ string {% function (d) { return this.toUpper(d[2]) } %}
+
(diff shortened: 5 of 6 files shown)


JacobEdelman:
This definitely seems rather useful. Hardmath123 is currently on vacation but he'll be back in a couple of weeks.

rwindelz:
@jlarsson I understand the wish to abstract the AST constructors; it can be quite handy.  Would you compare & contrast this idea of a context with the code blocks that Nearley already supports.
ie. 
`prod -> word whit? ("-"|"="):+ ">" whit? expression+  {% function(d) { return {name: d[0], rules: d[5]}; } %}
      | "@" whit? js  {% function(d) { return {body: d[2]}; } %}
`
In particular, the `"@" whit? js` which allows you to include arbitrary function definitions
eg.
`@ {% var id = function (d) { return d; } %}
`

jlarsson:
(_apologies for ranting_) @rwindelz Code blocks are unfortunately not comparable to fully externalized contexts.
* Code blocks are an implementation, while the context acts as an api (from the grammar perspective)
* Code blocks are maintained along with the grammar while contexts can reside anywhere and can benefit from unittests, linters and other JS centric methodologies
* Code blocks kind of often forces parsing results to have the form of the lowest common denominator, while contexts can be made arbitratily specific to the desired task. Lets use the calculator grammar as an example. It would require a substantial rewrite of that grammar to have a parsing return a syntax tree. On the other hand, its not hard to feed the parser with a AST factory (the context) which in one implementation performs direct evalutation and in another builds a syntax tree. The language (the grammar) stays the same, but the interpretation of sentences spoken in that language is left to application.

In OO terms, grammars can now adapt the Single Responsibilty Principle (i.e. it knows syntax but nothing else) and the Open Closed Principle (grammar stays fixed but added context implementations can produce new interesting results).
These are (for me) crucial values when development of a parser matures from a simple pipeline tool into a mission critical business value.

JacobEdelman:
I have to agree that there are many times where contexts of this type would have been useful.

rwindelz:
@jlarsson  We have different ideas on what is a rant :-)  I appreciate the clarity of your explanation.
Regarding your bullets.  On point 1 I am ambivalent, perhaps only because with some discipline using code blocks you may achieve something similar.  On point 2 & 3 I absolutely agree.

Many parser implementations clutter up the grammar specification with actions / semantics and while the semantics are required to do anything useful with the parse, the penalty with respect to the clarity of the grammar specification is high (too high in many cases - my personal feeling on this).

Out of general interest I keep an eye on the activities at http://vpri.org/ in particular the work done by Ian Piumiarta and Alessandro Warth ( @alexwarth ).  Of particular interest has been OMeta and now https://github.com/cdglabs/ohm.  Of note in Ohm is the notion that the grammar should be specified [without the semantics](https://github.com/cdglabs/ohm/blob/master/doc/philosophy.md).    The semantics in Ohm are injected later just as you are proposing here (for the same reasons).  I can recommend https://github.com/cdglabs/ohm/blob/master/examples/math/index.html where you should see striking similarities with your contexts.  OMeta & Ohm do have one deal breaker for me in that they are based on PEG with its ordered choice requirement.
I like the clean look of the Ohm grammars and the fact that the host language does not appear in the grammar specification.  Nearley with contexts still has a host language component in the grammar specification however it would be a small change to the grammar to do the equivalent of Ohm.  (at the same time I would drop code blocks as they are subsumed by contexts)

jlarsson:
Thanks for the reading references :)
As much as I strive for separation between grammar and semantics, my pragmatic angle is that code blocks and contexts are highly complementary.

If postprocessors where left only to simply delegate to context methods, code blocks are obviously redundant. But (again, this is my opinion based on my coding style), postprocessors are best when they can offer some utility - typically fixes to grammar ambiguity or just patching otherwise hard to express grammars. The examples that comes to my mind are 
* trivial reductions: reducing terminals before passing on to context (joining _[a-z]:+_ and such)
* hiding grammar structure: (programming) languages are noisy;  keywords, punctuations, whitespace and what else. The inlined postprocessor does the selection of which sub parsings to pass along to the context. This issue might be less relevant with _named tokens_ (#58), but the key point is that _semantics must be decoupled from actual grammar structure_.
* early rejection: my personal style would be to (at least try) reject early in inlined postprocessor code instead of delegating that decision to the context. I think. Its still a hypothetical question for me and the need for it might be an indication of ~~sloppy~~ unwanted side effects in semantic actions...

In Ohm (if I understood it correctly), the semantic part must be fully aware of every grammar detail, and you are essentially writing down your productions again, albeit in another form. I strongly object to this anti pattern (duplicate code). 

Ohm has one big advantage  though - semantic actions are invoked exactly once and never for rejected derivations. But that is a property of PEGs (which kills ambiguity) and not really a consequence of how Ohm evaluates semantics.

rwindelz:
re: my dropping code blocks comment
sorry, i mistakenly said 'drop code blocks' when i was thinking specifically of the code blocks that are not attached to any rule, ie, `"@" whit? js {% function(d) { return {body: d[2]}; } %}`

re: the anti pattern of duplicate code
It's not really duplicate code in that while grammar and semantics are isomorphic, one disassembles and the other reassembles.  Think of fold/unfold in algebraic data types; the grammar unfolds the input tokens into the abstract datatype and the semantic function folds the abstract datatype to produce the meaning value.   You have a choice of interleaving the specification or factoring apart.  If you allow for specification in either the grammar or the semantics then the reader may have to consult two separate documents to understand.

The Earley algorithm also has a property that for any specific start position in the input stream the semantic actions for any non terminal will only be invoked once with the caveat that for right recursive grammars a small change to the algorithm (that subsumes [Joop Leo's](http://www.sciencedirect.com/science/article/pii/030439759190180A)) is required.  The semantic actions will only be invoked once for each parse result returned.  This effectively memoizes intermediate parse results across ambiguous parses.  To defer evaluation, wrap results in lambdas.

cheers

Hardmath123:
Hi there! Thanks for taking a look at nearley.

I'm a little concerned that this doesn't actually add any functionality that nearley doesn't already have. You can do something like

```
@{% var ctx = require("context.js");  %}
# …
# …
a -> something something complicated grammar {% ctx.foo %}
```

As such, I'm not too worried about separating grammar from postprocessor functionality—nearley is a JavaScript-based parser and so it doesn't need to be language-agnostic.

jlarsson:
@Hardmath123 Binding to a module using `require` is entirely different (and strictly weaker in an OO perspective) from invoking actions in a truly dynamic/polymorphic context.
You are advocating globals, I'm more in for separation of concerns.

Hardmath123:
I'm not sure separation is the best choice in this case. It's really a question of what you want from your parser, and I don't think a parser should do too much. Postprocessors shouldn't have any complex logic in them; that should almost always be left up to the code that will interpret the AST. Part of the appeal of describing a language as syntax + semantics is that the two halves can be mutually agnostic.

Maybe it would be more convincing to me if you showed a use case where providing this "context" object would be substantially nicer?

jlarsson:
In unit testing or any other situation where I want to have explicit/special/different semantic interpretation of a parsing,

Hardmath123:
Sorry, lost track of this discussion for a bit because of school and stuff.

Rereading the discussion, it seems like an alternative solution, that has all of the benefits you described, would be to produce a "global" variable `$ctx` or something which contains all the "context". I think that would be a nicer solution because using `this` requires you to create a `{% function(d) {this.makeTree(d); } %}`, whereas with the global variable we can write post-processors as `{% $.makeTree %}`. Additionally, I would like to reserve the `this` binding for some other purpose in the future—maybe meta-information about the parse.

Finally, it feels weird having the mocha test for just this one feature. Could we just write a simple test-case in the `test/` directory, along with the remaining tests?

Thanks, and sorry again for the 1.6 month delay…

jcorbin:
Hmm, also if you're going to do anything to introduce more standardized testing, I'd recommend using https://github.com/substack/tape instead of mocha.

solmsted:
This is an older discussion, but for what it's worth, I really like the idea of being able to pass in some data that postprocessor functions can use.  I think it would work better if it was passed as an argument to the function instead of using `this`.

Hardmath123:
Mmm, okay, so this branch apparently has merge conflicts but let's see if either @jlarsson or I can get it up-to-date and merged for @solmsted to use. :-)

Oh, and I like the idea of it being an argument rather than `this`… it seems more natural to me. Perhaps we can make `this` point to the parser itself (if it doesn't already)?

nanalan:
For those that really need this feature, [nearley-make](https://github.com/nanalan/nearley-make#usage) supports passing arguments to your grammar (second argument to `make`). It's not a perfect solution, but it works :smile:

vietlq:
Any update here?

tjvr:
It's an interesting idea. It won't happen anytime soon but we'll think about it for v3. :-)

Hardmath123:
Closing this PR — I doubt this will be actually useful, and it has too many merge conflicts.

PR #69, opened 2015-09-12
Added preprocessor templates.

### LINKED ISSUE #67: JS preprocessors
I'm quite accustomed to CoffeeScript, so I like to use it here and there. It would be nice if I can use a custom preprocessor with nearley (e.g. Babel, CoffeeScript, 5to6, PromisedLand...).

bojidar-bg:
Closes #67 

This should work in browsers too (but it wasn't tested...).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules/*
 .DS_Store
 test/profile.log
 test/parens.js
+test/tmp.*
 *.swp
 *.ps1
 *.cmd
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -17,6 +17,7 @@ var Compile = function (structure, opts) {
 
     var outputRules = [];
     var body = []; // @directives list
+    var config = {};
 
     function buildProcessedRule(name, rule) {
         var tokenList = [];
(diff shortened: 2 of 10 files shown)


Hardmath123:
Hmm, I'm not sure I understand what's going on here. Why are there templates? I thought you just need a simple way to pass stuff in `{% %}` through the defined preprocessor.

bojidar-bg:
@Hardmath123 The problem is that some preprocessors with insert a blob of code on top (Babbel), or will enclose everything in a closure (CoffeeScript), so I though that is will be simpler if we just make an output "template" that will translate the rules, etc. for that preprocessor (and then the preprocessor can just do its magic).

Maybe it is also possible to pass the postprocessors through a preprocessor as you say, but it will be a different way for each preprocessor. It would be nice if we an just tell the preprocessor "This is pure javascript, don't touch it," but some of them don't even have such option (TypeScript, AFAIK).

I agree that templates are a bit overkill, but for the time being I can't see much alternatives.

Hardmath123:
Ah, I see. That makes sense; thanks for explaining.

I'm still hesitant to include ejs: it's heavy, and I don't want to risk breaking browser support. Is there a way to rewrite the parts that use ejs manually, as it is written right now? I think that will make the code much smaller.

Also, could you test it out in the browser? (Both generation of grammars and execution of grammars.) I do care a lot about browser support. :-)

bojidar-bg:
@Hardmath123 ok... that makes sense too :wink:. I think I can rewrite them, np :smile: (Maybe next Saturday, who knows...)

Hardmath123:
Thanks @bojidar-bg ! I'm really excited about this feature. Someday I could be writing postprocessors in ClojureScript. :D

bojidar-bg:
@Hardmath123 it will be more interesting if it is possible to write them in two preprocessors, e.g. some in CoffeeScript and some in ClojureScript, but... that's another story ^^

PR #70, opened 2015-09-20
Better named sub rules

jcorbin:
This makes it easier to follow generated code from the compiler.

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -32,7 +32,7 @@ var Compile = function (structure, opts) {
                         };
                     });
 
-                    var newname = unique(" string");
+                    var newname = unique(ruleName + "$string");
                     buildProcessedRule(newname, {
                         tokens: rules,
                         postprocess: joiner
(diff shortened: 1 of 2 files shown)


jcorbin:
Also re-bootstrapped the language for a demo of why this is easier to read.

jcorbin:
Not sure if you want `test/{ebnf,negatable}.js` also re-genned.

jcorbin:
Note: this PR is the tip of a larger working branch where I'm coming to terms with your compiler: https://github.com/Hardmath123/nearley/compare/Hardmath123:master...jcorbin:better_compile

Let me know how interested you are in further improvements here, I can start decomposing that branch into more digestable stages in subsequent PRs.

Hardmath123:
Very nice. This was one of those little hacks that bothered me but I never had motivation to fix… thanks for doing this.

I would love more improvements to the core compiler. And yeah, it would be *amazing* if you could break it up into separate easy-to-review PR's, but if that's too much work, then don't worry too much about it. :-)

PR #72, opened 2015-09-20
Module only, no window

jcorbin:
So here's another idea: rather than having a bunch of manual module.exports vs window logic, how about a bundler.

This probably needs a bit more work, and is more of an idea / conversation starter PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ To use a generated grammar in a node runtime, install `nearley` as a module:
 To use a generated grammar in a browser runtime, include `nearley.js` (you can
 hardlink from Github if you want):
 
-    <script src="https://raw.githubusercontent.com/Hardmath123/nearley/master/lib/nearley.js"></script>
+    <script src="https://raw.githubusercontent.com/Hardmath123/nearley/master/nearley.js"></script>
     <script src="my-generated-grammar.js"></script>
 
 
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -1,9 +1,7 @@
-(function () {
-if (typeof(window) === 'undefined') {
-    var nearley = require('../lib/nearley.js');
-} else {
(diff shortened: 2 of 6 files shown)


jcorbin:
One issue: I've done a commit of the browser artifact to master, but that's generally considered an anti pattern.

It's a bit nicer to have a separate branch for publishing, say using gh-pages, and to only commit artifacts to it.  One advantage of this is that you avoid duplication in the npm package, which has no use for browser concerns (also why browserify is a dev dependency here, not a runtime/production one).

For an example of how I've most recently solved this, see https://github.com/jcorbin/hexant, but note that there my context is a full single-page app and not a library... (at least yet...).  Also as a fun aside, this is the project that I'll be using earley in ;-).

Hardmath123:
Hmm, I'm always kind of hesitant to add dependencies like browserify. As it stands, only `nearley.js` is required to actually parse stuff; `compile` and `generate` are only useful for the (yet-to-be-properly-implemented) online demo compiler.

I'm curious, how does an Earley parser help build a cellular automata simulator? :-)

jcorbin:
Well in this case browserify is only a dev dependency, it's only needed to build new browser bundles.  This one's just a crazy idea tho, so feel free to close out if you'd rather just say no for now :-)

On the hexant question, see:
- here's my current "parser" https://github.com/jcorbin/hexant/blob/master/turmite.js#L99 that only supports ants
- here's my WIP towards building a grammar for all of the turmite functionality lurking in 0.7 hexant: https://github.com/jcorbin/hexant/blob/dev/turmite.ne

It's going to be a mini rules language, e.g. an ant should look something like:
```
T = turn(L R)
0, c -> 0, c+1, T[c]
```

Hardmath123:
I don't think this is going to happen anytime soon — closing the PR.

PR #71, opened 2015-09-20
Better compile prep

jcorbin:
Superficial prep work out ahead of compiler embetterment.

PR #73, opened 2015-09-21
Fix generate indent

jcorbin:
As discussed on #71, this fixes `lib/generate.js` so that it outputs more consistent output.

The churn in `lib/nearly-language-bootstrapped.js` is due to now correctly stripping leading space from function#toString output.


Hardmath123:
Shiny.

I think the `dedent` solution is very clever, but I think it might fail in a couple of pathological cases such as

```
a -> "foo" {%
function (x) {
    return x;
             }
%}
```

Maybe you want to "dedent" by the minimum number of leading spaces in `lines[1:]` (Python slice syntax?) rather than the number of leading spaces in the last line, which is how I understand it works right now.

If you think that's messy, though, I'm happy to merge as-is. Your call.

jcorbin:
Good call, changed to take min indent over the tail; verify no change under `npm run bootstrap`.

Hardmath123:
One of the lines in the diff seemed duplicated; is this a typo or am I missing something?

jcorbin:
Fixed the derplicated line.

Hardmath123:
Nice, merged. Thanks. :-)

PR #74, opened 2015-09-21
The Great Compile Breakup

jcorbin:
Breaks each token builder out, makes rule and token building more functional.

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -23,10 +23,7 @@ function Compile(structure, opts) {
                 result.body.push(productionRule.body);
             }
         } else {
-            var rules = productionRule.rules;
-            rules.forEach(function(rule) {
-                buildProcessedRule(productionRule.name, rule);
-            });
+            produceRules(productionRule.name, productionRule.rules);
             if (!result.start) {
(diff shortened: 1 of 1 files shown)


jcorbin:
I ended up putting more into this PR than I first thought I would, let me know if you'd rather this one be dialed back more.

I strongly suggest ignoring whitespace to review this PR (add `?w=1` to github, or just use git with `-w` git option).  You may also find it easier to review commit a time.

Hardmath123:
Looks good, but I won't be able to review/merge things until the week-end because of school and stuff. :-)

jcorbin:
Weekend bump ;-)

Hardmath123:
Very nice, thanks so much.

Out of curiosity, why did you prefer `for` instead of `[].forEach` in 9ab912e? Is there a performance issue, or is it more of a stylistic choice?

jcorbin:
Yeah, for loops are indeed substantially faster than a forEach when it matters.  Also trying to use a network of high level functions (forEach, map, reduce, filter, etc) tends to be much harder to understand in larger combination than simple for loops and such.

PR #75, opened 2015-09-22
Added preprocessor support.

### LINKED ISSUE #67: JS preprocessors

bojidar-bg:
Closes #67

This time it should be better than last time :smile:.

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -83,10 +83,15 @@ function serializeRule(rule, builtinPostprocessors) {
 }
 
 var generate = function (parser, exportName) {
-    return (generate[parser.config.preprocessor] ?
-      generate[parser.config.preprocessor] :
-      generate["_default"]
-    )(parser, exportName);
+    if(!parser.config.preprocessor) {
+      parser.config.preprocessor = "_default";
+    }
+
+    if(!generate[parser.config.preprocessor]) {
+      throw new Error("No such preprocessor: " + parser.config.preprocessor)
+    }
+
+    return generate[parser.config.preprocessor](parser, exportName);
 };
 
 generate.js = generate._default = generate.javascript = function (parser, exportName) {
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -1,8 +1,8 @@
 (function () {
 
-function serializeRules(rules, specialPostprocessors) {
+function serializeRules(rules, builtinPostprocessors) {
     return "[\n    " + rules.map(function(rule) {
-      return serializeRule(rule, specialPostprocessors);
+      return serializeRule(rule, builtinPostprocessors);
     }).join(",\n    ") + "\n]";
 }
 
@@ -64,13 +64,13 @@ function id(a) {
   return a;
 }
 
-function serializeRule(rule, specialPostprocessors) {
+function serializeRule(rule, builtinPostprocessors) {
     var ret = '{';
     ret += '"name": ' + JSON.stringify(rule.name);
     ret += ', "symbols": [' + rule.symbols.map(serializeSymbol).join(', ') + ']';
     if (rule.postprocess) {
-        if(rule.postprocess.special) {
-            rule.postprocess = specialPostprocessors[rule.postprocess.special];
+        if(rule.postprocess.builtin) {
+            rule.postprocess = builtinPostprocessors[rule.postprocess.builtin];
         }
         ret += ', "postprocess": ' + tabulateString(dedentFunc(rule.postprocess), '    ', id);
     }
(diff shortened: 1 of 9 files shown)


bojidar-bg:
@Hardmath123 ping?

Hardmath123:
Yeah, I saw this. Looks good, but I can't review/merge until the week-end. :-)

bojidar-bg:
:smile: Ok, I can wait a few days ...

bojidar-bg:
Note: This PR is a bit broken, as some ebnf functions get serialized as JS not CoffeeScript

bojidar-bg:
Ok, no longer broken, fixed it (and made a small merge conflict in the process :smiley: )...

Hardmath123:
Looking over this right now… could you explain what `tabulateString` does?

Also, I believe CoffeeScript allows you to include literal JavaScript in backticks or something. Then you don't need to rewrite built-in postprocessors in CoffeeScript. The whole `{"special": x}` seems a little icky and best avoided if possible (minimizing the number of moving parts and all that).

Thanks, and sorry again for the wait. :-)

Hardmath123:
Oh, one last thing: after @jcorbin's refactoring, this PR now has some merge conflicts. Shouldn't be too hard to fix, though.

bojidar-bg:
@Hardmath123 About the backticks, it isn't modular enough, e.g. TypeScript won't allow it (who know for ClosureScript, etc.)

bojidar-bg:
`tabulateString` will add a string to every line of the an input string (or, if it is an array, it will first join it with "\n")

I decided to go with `{"special": x}`, because it allows for every generator function pick it's own. (The other option was to use some special string, e.g. "___INTERNAL__DONT_USE___x___", but it seemed to much of a hack). Maybe we should change special to something else though...

bojidar-bg:
@Hardmath123 Ok, merged, now it is ready.

lib/generate.js
```diff
@@ -35,36 +37,64 @@ function dedentFunc(func) {
     });
 }
 
+function tabulateString(string, indent, filter) {
+    if(Array.isArray(string)) {
+      string = string.join('\n');
+    }
+    tabulated = string.toString().split('\n').map(function addIndent(line, i) {
```
Hardmath123: Seems inefficient, you're `join()`ing the array and then `split()`ing it again. :-/

lib/generate.js
```diff
@@ -35,36 +37,64 @@ function dedentFunc(func) {
     });
 }
 
+function tabulateString(string, indent, filter) {
```
Hardmath123: Do you ever actually use the `filter` argument? I don't see it being used anywhere except with `id`, which shouldn't change anything.

lib/generate.js
```diff
@@ -35,36 +37,64 @@ function dedentFunc(func) {
...
+    }).join('\n');
+    return tabulated;
+}
+
 function serializeSymbol(s) {
     return (s instanceof RegExp) ? s.toString()
                                  : JSON.stringify(s);
 }
 
+function id(a) {
```
Hardmath123: See comment about "filter", above.

lib/generate.js
```diff
@@ -35,36 +37,64 @@ function dedentFunc(func) {
...
     if (rule.postprocess) {
-        var lines = dedentFunc(rule.postprocess);
-        ret += ', "postprocess": ' + lines.map(function indent(line, i) {
-            if (i > 0) {
-                line = '    ' + line;
-            }
-            return line;
-        }).join('\n');
+        if(rule.postprocess.special) {
+            rule.postprocess = this.specialPostprocessors[rule.postprocess.special];
```
Hardmath123: I think it would be better to have `specialPostprocessors` as an argument rather than something bound to `this`.

lib/generate.js
```diff
@@ -1,7 +1,9 @@
 (function () {
 
-function serializeRules(rules) {
-    return "[\n    " + rules.map(serializeRule).join(",\n    ") + "\n]";
+function serializeRules(rules, specialPostprocessors) {
+    return "[\n    " + rules.map(serializeRule.bind({
```
Hardmath123: See comment on `:73`.

Hardmath123:
I left some thoughts as line notes.

lib/generate.js
bojidar-bg: the filter argument is passed the id of the current line along with the line itself. `id` will return 0 for id 0, thus preventing the indentation of the first line.

lib/generate.js
bojidar-bg: It's a bit hard with `map` on L4. But yeah, there should be a better way :smile:

bojidar-bg:
@Hardmath123 Ok, I'll make another commit it seems....

bojidar-bg:
@Hardmath123 Ok, fixed all comments, I think. Anything else?

Hardmath123:
Sorry, one last thing that bothers me: rather than a `_default`, an error should be thrown if no such preprocessor exists. So a grammar with `@preprocessor jkhldsfghjklfsg` should not compile. Just `throw new Error("No such preprocessor: "+…);` should suffice.

Everything else looks great. Thanks for including a nice test case. :-)

bojidar-bg:
@Hardmath123 Ok, now I think everything is ready!

Hardmath123:
Thanks; merged! :-)

PR #76, opened 2015-09-27
Clearer edge cases

jcorbin:
Okay, a few things going on here:
- explicit null rather than undefined for skipping tokens in a rule
- which allows us to then pull apart the if/else ladder into a set of decoupled short-circuiting ifs
- which then allows us to easily re-order the checks, and expand them in future changes
- use said flexibility to do the inexpensive "atomic" checks first

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -68,18 +68,26 @@ function Compile(structure, opts) {
             } else {
                 return undefined;
             }
-        } else if (token.subexpression) {
+        }
+
+        if (token.subexpression) {
             return buildSubExpressionToken(ruleName, token);
-        } else if (token.ebnf) {
+        }
+
+        if (token.ebnf) {
             return buildEBNFToken(ruleName, token);
-        } else if (typeof(token) === 'string') {
+        }
+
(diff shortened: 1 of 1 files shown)


lib/compile.js
```diff
@@ -57,26 +57,37 @@ function Compile(structure, opts) {
...
+
         if (token.literal) {
-            var str = token.literal;
-            if (str.length > 1) {
-                return buildStringToken(ruleName, token);
-            } else if (str.length === 1) {
-                return {
-                    literal: str
-                };
+            if (!token.literal ||
```
Hardmath123: A little confused by this check. Is this trying to check for the empty string? If so, wouldn't `token.literal === ''` suffice?

Hardmath123:
(In case you missed it, see line note on :72).

lib/compile.js
jcorbin: Oh hmm, it's wholly redundant with L71 anyhow, will just remove it.

jcorbin:
Updated

Hardmath123:
Thanks, merged. :-)

PR #84, opened 2016-01-08
Added a new example for parsing numbers as strings to regular numbers

jaukia:
A nearley.js example for parsing strings like "two million seven hundred four" to numbers. Feel free to include this in the examples, if you like. This would be fairly easy to extend to handle strings with calculations, such as "two million seven hundred four plus one hundred seventy five".

### DIFF
--- a/examples/js/numberstring.js
+++ b/examples/js/numberstring.js
@@ -1,0 +1,154 @@
+// Generated automatically by nearley
+// http://github.com/Hardmath123/nearley
+(function () {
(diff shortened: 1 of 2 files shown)


Hardmath123:
I love it! This is such a cool idea. Merging now. Thank you so much. :-)

Hardmath123:
```
$ nearley-unparse examples/js/numberstring.js 
seventy one
```

Haha. This is so much fun to play with.

Issue #85, opened 2016-01-22
Spurious parses stemming from nullable nonterminals


corwin-of-amber:
The following grammar:
```
@builtin "whitespace.ne"

d -> a

a -> b _ "&"
   | b

b -> letter
   | "(" _ d _ ")"

letter -> [a-z]
```
when run using nearley-test on ```(x)```, generates two (identical) parses ```[ [ [ '(', null, [ [ [ [ 'x' ] ] ] ], null, ')' ] ] ]```. Since the above grammar is unambiguous, this is unexpected.

Notice that this does not occur if you omit the rule ```a -> b _ "&"```, which does not appear in the derivation, and this is even more unexpected. It has to do with the order of prediction.

I am preparing a pull request that suggests a fix.

corwin-of-amber:
Tried 2.2.0 and it works :+1:

Hardmath123:
Awesome. Closing, then.

PR #86, opened 2016-01-22
Proposed fix for Issue #85

corwin-of-amber:
The problem occurs because the loop in nearley.js:108 goes over all the states before the currently processed rule. However some of the states in range 0...ind, although having been inserted before the current state was inserted, may still have been *processed after* the the current state was inserted. This will result in double consumption of the nullable nonterminal.


Hardmath123:
Hi @corwin-of-amber ! Thanks for catching this bug. Those nullables are… gnarly. Very, very touchy.

I'm always pretty cautious about merging PRs that modify `nearley.js` itself, so I expect to review this for a couple of days before I'm convinced that nothing else broke (as I said, nullables are touchy). But you look like you know what you're doing. :-)

P.S. Maybe put `push_epsilon` in `camelCase` instead? Seems to fit `nearley.js` better (sorry; this is picky on my part!).

corwin-of-amber:
Thanks for your quick response!

I guess I was a bit sloppy on the conventions here. Actually, I believe that push_epsilon might go better in State.prototype rather than on the module level. I will try a few alternatives and see how it looks.

Also, to get over the fear of changing the runtime, and if you approve, I am willing to make a small test suite, using chai or similar, out of the test cases already in the test/ subdirectory and perhaps a few more textbook examples. In any case, thanks for developing and maintaining nearley, it’s a great thing to have Earley support in node!

corwin-of-amber:
Pushed an initial version of a test file.
To run (in case someone is reading this who has never used Mocha before):
```
npm install -g mocha
mocha test/launch.js
```

Hardmath123:
That test file looks nice. I haven't used chai/mocha before… it's hard to tell the difference between them. Looks like mocha "runs" the tests while chai is the DSL used to describe them, by modifying prototypes of built-in objects? Hmm. Requires more research on my part.

If you put this test file in a separate branch/separate PR, I'd be happy to merge it right now so that the tests are not waiting on a review of the epsilon fix. :-)

Thanks again for your contribution!

Hardmath123:
(FYI I've resolved all merge conflicts on my local copy, so don't worry about that. Currently reviewing.)

Hardmath123:
Merged by latest push. Thank you so much for this bug fix. :-)

PR #87, opened 2016-01-24
Mocha tests

corwin-of-amber:
Here are some tests using Mocha & Chai, in a separate branch so they can be merge without worrying about changes to the runtime. Currently the test "nullable whitespace bug" fails (fix in PR #86).


Hardmath123:
Nice. Merging. Thank you. :-)

PR #92, opened 2016-03-19
fix npm test error on windows

steambap:
fix 'npm test' error ( test on windows 10 / node v5.8.0 & Linux 4.2.0-c9 / node v4.4.0 ) 

1.on windows you have to say 'node XX.js' instead of only 'XX.js'
2.to make test 'should build for CoffeeScript' pass, I add coffee script as dev dependency (I don't think I everybody install coffee script on their machine in 2016)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "chai": "^3.4.1",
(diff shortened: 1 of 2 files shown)


Hardmath123:
Thanks for the PR!

PR #93, opened 2016-03-23
add build.js to make build process more portable

steambap:
The new build.js file behave exactly the same as build.sh except it does not run the test (test / launch.sh).

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,0 +1,6 @@
+require('shelljs/global');
+
+echo('Building nearley...');
+exec('node bin/nearleyc.js lib/nearley-language-bootstrapped.ne > grammar.tmp');
+echo('Deleting temp file...');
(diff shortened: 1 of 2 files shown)


Hardmath123:
Thanks for this! I'm going to hold off on doing any npm-ey stuff for a bit due to the weird environment right now but I'll probably pull this this weekend.

steambap:
what happened to your npm environment?

Hardmath123:
You might have heard, npm's going through a bit of a crisis right now since a package with a lot of dependents was unpublished suddenly and a lot of stuff broke downstream. I don't want to push any changes in the middle of this chaos. :-)

Hardmath123:
I'm not sure about merging this; the shelljs module seems like overkill. If anyone is actually developing for nearley on Windows, they can probably run the build commands manually or write a batch script.

PR #99, opened 2016-05-19
add positive decimals to builtins

seiyria:
there are positive integers, I think positive decimals should be added for consistency

### DIFF
--- a/builtin/number.ne
+++ b/builtin/number.ne
@@ -14,6 +14,14 @@ int -> ("-"|"+"):? [0-9]:+ {%
     }
 %}
 
+posdecimal -> [0-9]:+ ("." [0-9]:+):? {%
+    function(d) {
+        return parseFloat(
+            d[0].join("") +
+            (d[1] ? "."+d[1][1].join("") : "")
+        );
+    }
+%}
 
 decimal -> "-":? [0-9]:+ ("." [0-9]:+):? {%
     function(d) {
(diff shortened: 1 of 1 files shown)


Hardmath123:
Mmm, interesting. See, looking at this more carefully, it makes sense that what is currently `int` should actually be called `unsigned_int`, and `posdecimal` should actually be called `unsigned_decimal`.

So, two questions.

1) Do you have an actual use-case for `unsigned_decimal`, or is it literally just for consistency?
2) Would you be willing to update the PR, renaming `pos_…` to `unsigned_…`?

Thanks!

seiyria:
I do have a use case. In my project, my language has a few variables pulled
out that represent offsets that can't be negative.

I'd be willing to update the pr. I can do it tomorrow if you want to merge
it. Will you also be updating the int to match with that release?

Hardmath123:
Yeah, updating int would be nice as well.

Thanks!

seiyria:
@Hardmath123 done! :D

bobbybee:
@seiyria I suspect you'll run into bugs anyway if you drop a variable for the offset... which you should support..

seiyria:
@bobbybee I'm not sure I follow what you mean. With the language I'm building, variables only exist as part of loops, and when using a loop you can only iterate from 0 to a number. It's intentionally restrictive and should be manageable.

Hardmath123:
Thanks @seiyria ! This is great; I've merged it into master, and you'll be able to use it in the next release of nearley (or you could start using it now by installing nearley from the git repository).

seiyria:
Awesome! When will you be releasing the next version?

Hardmath123:
It should be online now.

seiyria:
Great, thanks again!

PR #107, opened 2016-07-20
add nearley-make as nearleyc alternative

nanalan:
I've created a package purely for compiling _and_ running a `grammar.ne` file called [nearley-make](https://github.com/nanalan/nearley-make).

It works inside JS runtimes only, rather than the shell. nearley-make takes a grammar source string as input and outputs an instance of `nearley.Parser`, so it's mostly for development purposes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ To compile a parser, use the `nearleyc` command:
 
 Run `nearleyc --help` for more options.
 
+There is also a version of `nearleyc` for node runtimes, [nearley-make](https://github.com/nanalan/nearley-make).
+
 To use a generated grammar in a node runtime, install `nearley` as a module:
 
     npm install nearley


Hardmath123:
Feedback on the project: Hmm, looks cool but also kind of a hack. I mean, isn't it easy enough to just run nearleyc through a shell? There are dozens of ways to implement automatic compilation of source files (watch, compile-on-write, …). I guess I'm looking for a specific use-case for this, and not finding one—perhaps you have one in mind? Also, http://stackoverflow.com/a/17585470 could help avoid your string-replacement kludge.

Feedback on the PR: I think the sentence "There is also a version of `nearleyc` for node runtimes, nearley-make" is just confusing for a user, creating many unanswered questions. For this reason, I'm hesitant to merge this PR.

nanalan:
> Hmm, looks cool but also kind of a hack.
Agreed. Thanks for the StackOverflow link!

> I guess I'm looking for a specific use-case for this, and not finding one—perhaps you have one in mind?
A specific use-case would include [nanalang](https://github.com/nanalan/g); if you peer into the source there is a version of it used with the `--dev` flag, so one doesn't have to compile the Nearley grammar before interpreting themselves.

It's an edge-case, but still helpful :smile:

Hardmath123:
> …so one doesn't have to compile the Nearley grammar before interpreting themselves.

I see, so you just want to avoid recompiling the grammar each time. You should look into tools like `watch`, `inotifywait`, `make`, or editor plugins that automate this process. I'm definitely not going to merge tools that encourage this kind of usage of nearley, at least not until they're more mature and have actual use-cases.

nanalan:
Sure. :package:

###### (sometimes build tools aren't the answer, but alright)

PR #109, opened 2016-07-21
Fix regression in number.ne

rahatarmanahmed:
The original fix was added in ba4cebc86aa5b430f4574a1a08b106a49816c5d3, but was lost in e552c8839be38514e53737667a239bc229344bc2.


Hardmath123:
Aww, gah, I hate bugs like this—embarrassing and hard to catch! One of these days I should definitely take the time to put in tests for the builtin libraries.

Thanks so much for the fix.

PR #112, opened 2016-07-25
Performance optimization:

bd82:
"findLeo" seems to take 80% of the runtime for a JSON grammar example.

* Avoids iterating over the row twice.
* Avoids the creation of the anonymous function in the map/filter
  high order functions.

Results in Chrome (V8).
X3 speed in JSON Grammar + Tokenizer.
x2 speed in JSON Grammar without Tokenizer.

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -71,11 +71,15 @@ State.prototype.process = function(location, ind, table, rules, addedRules) {
         if (!(this.data === Parser.fail)) {
             // LEO THE LION SAYS GER
             function findLeo(idx, rulename, finalData) {
-                var items = table[idx].map(function(s) {
-                    return s.consumeNonTerminal(rulename);
-                }).filter(function(s) {
-                    return s && s.isComplete() && s.rule.name === rulename;
-                });
+                // performance optimization, avoid high order functions(map/filter) in hotspot code.
+                var items = [];
+                var row = table[idx];
+                for (var col = 0; col < row.length; col++) {
+                    var s = row[col].consumeNonTerminal(rulename);
+                    if (s && s.isComplete() && s.rule.name === rulename) {
+                        items.push(s);
+                    }
+                }
                 if (items.length === 1) {
                     var item = items[0];
                     item.data[item.data.length-1] = finalData;


Hardmath123:
Haha, see, this is exactly the kind of thing I would never think to do. I come from a Scheme background where that sort of thing is considered an abomination… :-)

Thank you for the optimization. I'm sure this kind of error is present all over nearley's source.

bd82:
It is indeed an abomination :cry: .
But not an error. Only the hot spots that need to be optimized like that.
It's best that as much of the code as possible will stay in a functional style for maintainability.
(and sanity :smile: )

Here is an example of a super optimized RegExp based Lexer Tokenize method:
https://github.com/SAP/chevrotain/blob/master/src/scan/lexer_public.ts#L250
It is all **for** and **while**. no anonymous functions, no recursion, not even splitting to sub functions (~200 lines!) to avoid unnecessary function invocations.

On the other hand, code that only runs **once** on the initialization of that same Lexer engine instances and thus is not a hot spot. is using a much more functional style:
https://github.com/SAP/chevrotain/blob/master/src/scan/lexer.ts#L41

There are no absolutes, the most suitable tool/paradigm for the job should be chosen.

bobbybee:
@bd82 To be fair, a lot of this mess can be avoided with a suitable JavaScript -> JavaScript compiler :-)

bd82:
@bobbybee Such as?

Does Babel have a plugin that transforms functional code to imperative code?
Or are you referring to some other transpiler ?

bobbybee:
@bd82 It's been a while, but I imagine Closure Compiler would do this if it's such an important optimization, no?

bd82:
Lets assume that it does do so for **built-in** JavaScript high order functions (map/filter/forEach).
That is easier, because their semantics are known (part of ECMA spec). although identifying that the the built-in functions are used could still be problematic.

Example:
```javascript
var hasNativeLike = {
  map: function() { return 666},
  filter:function() { return this.map()}
}
var canBeOptimized = [1,2,3,4].map(...).filter(...)
var shouldNotBeOptimized = hasNativeLike.filter(...)
```

But would it do so for the dozens of custom methods commonly used by a projects like [Lodash](https://lodash.com/docs)?  or [Ramda](http://ramdajs.com/docs/)?

```javascript
var _ = require("lodash");
var result = _.dropWhile(_.reject([1,2,3], (x) => x % 2 ), (y)=> y >1)
```

In that case the semantics are not known in advance thus performing the optimization is harder.

I don't know what it can really do but from
the little details I managed to find about Google's closure compiler is that when advanced optimizations are active it performs dead code elimination and function inlining (in some cases).
not complete code transformations.

PR #113, opened 2016-07-28
Storing a single item in findLeo instead of a list.

JacobEdelman:
In findLeo a whole list was being used to store items when all that was needed was to check whether their exists one and only one item in the list. I'm unsure of how I should be checking this for changes in overall speed.

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -72,16 +72,20 @@ State.prototype.process = function(location, ind, table, rules, addedRules) {
             // LEO THE LION SAYS GER
             function findLeo(idx, rulename, finalData) {
                 // performance optimization, avoid high order functions(map/filter) in hotspot code.
-                var items = [];
+                var item = null;
                 var row = table[idx];
                 for (var col = 0; col < row.length; col++) {
                     var s = row[col].consumeNonTerminal(rulename);
                     if (s && s.isComplete() && s.rule.name === rulename) {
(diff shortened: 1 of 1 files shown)


bd82:
Hi.

There is a benchmark of a JSON grammar here:
http://sap.github.io/chevrotain/performance/
In this benchmark findLeo uses ~80% of the cpu time.

you can try to run this locally with your modified file by:
* Checking out the [source code](https://github.com/sap/chevrotain/tree/gh-pages/performance)
* Modifying the [script tag]( https://github.com/sap/chevrotain/tree/gh-pages/performance) that loads Nearley to point to your own local modified version.
* open the index.html in the browser you want to test in.

Hardmath123:
Please benchmark this—Shahar's works fine.

If you want to be really helpful: The "correct" way to solve this is to cache Leo items so you don't have to go down the rabbit hole each time you right-recurse.

JacobEdelman:
I got no significant difference with the benchmark. Can you explain the caching (or, if you don't have the time, feel free to merge what I have now).

Hardmath123:
Okay. I'll take a look and merge it eventually.

Don't worry about the caching; I'll get it implemented correctly when I have time.

Hardmath123:
Too many conflicts, and also the Leo code has changed in the meantime. Closing PR for now.

PR #119, opened 2016-08-25
Improve HTML output of railroad, including a meta-tag for UTF-8

cameronhunter:
Improving the HTML document produced by `nearley-railroad`. The document now specifies a UTF-8 charset allowing for unicode characters within railroad diagrams to be rendered correctly.

### DIFF
--- a/bin/nearley-railroad.js
+++ b/bin/nearley-railroad.js
@@ -102,6 +102,7 @@ function railroad(grm) {
     }
 
     return [
+      '<!DOCTYPE html>',
       '<html>',
         '<head>',
           '<meta charset="UTF-8">',
--- a/bin/nearley-railroad.js
+++ b/bin/nearley-railroad.js
@@ -45,20 +45,21 @@ function railroad(grm) {
         }
     });
 
-    ret = '<style type="text/css">\n';
-    ret += fs.readFileSync(
+    var style = fs.readFileSync(
         path.join(
             path.dirname(require.resolve('railroad-diagrams')),
             'railroad-diagrams.css'
         )
-    ).toString();
-    ret += '\n</style>';
+    );
 
-    Object.keys(rules).forEach(function(r) {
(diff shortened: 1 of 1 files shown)


Hardmath123:
Cameron — thank you for the nice, concise PR.

As it stands, nearley-railroad isn't meant to create standalone HTML documents, but rather snippets of HTML that in theory get copy-pasted into a "real" website that does have legitimate meta tags and whatnot. This is mainly because I guessed that people would almost certainly want their railroad diagrams embedded into text (for example, in a blog post) but also because nearley-railroad is still experimental in the sense that I have yet to meet a serious user who isn't me!

So the issue you bring up is really more general: should nearley-railroad output a full HTML document, or a snippet? Or something else entirely: maybe a PNG? I am open to your suggestions, but I would definitely like to think this through a little bit more before merging.

Thanks again.

bobbybee:
@Hardmath123 If you do choose to output a PNG (e.g.: via PhantomJS) or choose to embed the diagram by means of an iframe or equivalent, this PR might still be relevant. Else, if it's not feature creep (which I don't think it is), it can't hurt to have it as an option.

Whether it's enabled by default is perhaps a better question then.

cameronhunter:
I guess if it's for copy-pasting then there's not much difference for the user? They can still copy and paste the necessary parts. My only intention with the PR was so that unicode characters in my nonterminals would render correctly in the railroad diagrams 😃

Hardmath123:
Okay, for now I think it's okay for nearley to export standalone pages that can be included with iframes. The encoding issue definitely takes priority over my qualms about embedding (though, to be completely honest, I'm pleasantly surprised that special characters work in nonterminals).

Before I merge, can you make it print `<!DOCTYPE html>` at the top of the output?

Thanks.

Hardmath123:
Nice, thank you!

Issue #104, opened 2016-06-23
Unnecessary "new" in nearleyc


moll:
Hey again,

Noticed an unnecessary `new` on https://github.com/Hardmath123/nearley/blob/master/bin/nearleyc.js#L42.

Hardmath123:
Mmmm, weird, thanks for the tip. Would you like to PR a fix?

moll:
I wish I could, but I have hard enough of a time to get around to the plethora of my own OSS. Suggesting things through GitHub issues can do. :D

nanalan:
This can be closed as #120 was merged.

PR #120, opened 2016-08-26
Remove unnecessary `new` in require statement

cameronhunter:
As suggested in issue #104. Also removing unnecessary `.js` suffixes.

### DIFF
--- a/bin/nearleyc.js
+++ b/bin/nearleyc.js
@@ -1,10 +1,10 @@
 #!/usr/bin/env node
 
 var fs = require('fs');
-var nearley = require('../lib/nearley.js');
+var nearley = require('../lib/nearley');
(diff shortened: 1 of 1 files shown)


Hardmath123:
I don't see why we're omitting the `.js` — it doesn't make sense to me, and I'd rather keep the `.js` extensions unless there's an alarmingly urgent reason to omit them.

I'll merge just the `new` change if you PR that on its own.

cameronhunter:
For context, I removed the `.js` suffixes because it isn't normal convention to specify it.

Hardmath123:
Hmm, I'm not sure how I feel about that. I think I'll keep the `.js` extensions if that's alright with everyone.

Hardmath123:
\o/ Thanks!

PR #125, opened 2016-09-22
Which language is the comment comment about?

kanef:
Any time you're describing a language for describing a language, it can get confusing.  I found this line in the `README` confusing, because I didn't know whether it meant "You can add a nearley file, `#` starts a comment" or "This grammar defines a language that ignores comments beginning with `#`.  There's no use of `#` in the example that clarifies it.  Based on http://nearley.js.org/www/railroad-demo.html, I'm sure you meant the former, so how about describe it with a second example?

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,15 @@ expansions.
 Strings are the *terminals*, which match those string literals (specified as
 JSON-compatible strings).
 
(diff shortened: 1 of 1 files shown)


Hardmath123:
Thanks! I would never have guessed that line was confusing, so it's valuable to get this kind of feedback.

PR #130, opened 2016-10-01
Add --quiet flag to nearley-test.js

sandiegoscott:
As suggested.

### DIFF
--- a/bin/nearley-test.js
+++ b/bin/nearley-test.js
@@ -28,6 +28,11 @@ var opts = nomnom
         abbr: 'o',
         help: "File to output to (defaults to stdout)",
     })
+    .option('quiet', {
+        abbr: 'q',
+        flag: true,
+        help: "Output parse table only omitting parsing steps",
+    })
     .option('version', {
         abbr: 'v',
         flag: true,
@@ -57,7 +62,11 @@ var writeTable = function (writeStream, parser) {
                     writeStream.write(stateNumber++ + ": " + state.toString() + "\n");
                 } )
         } )
-    writeStream.write("\n\nParse results: \n");
+    writeStream.write("\n");
+}
+
+var writeResults = function (writeStream, parser) {
+    writeStream.write("\nParse results: \n");
     writeStream.write(require('util').inspect(parser.results, {colors: true, depth: null}));
     writeStream.write("\n");
 }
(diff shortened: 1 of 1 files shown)


bin/nearley-test.js
```diff
@@ -28,6 +28,11 @@ var opts = nomnom
         abbr: 'o',
         help: "File to output to (defaults to stdout)",
     })
+    .option('quiet', {
+        abbr: 'q',
+        flag: true,
+        help: "Output parse table only omitting parsing steps",
```
Hardmath123: We should change this helptext. The parse table *is* the thing you call "parsing steps". Try `"Output parse results only (hide Earley table)"`.

Hardmath123:
See my comment about the helptext. Other than that, it looks good.

sandiegoscott:
I'm new to PR's. I pushed the new help phrase in the 'quiet' branch, which now has 2 commits and also in a new 'quiet2' branch, with just one commit. Take your pick. Thanks.

Hardmath123:
Actually, you know what would be nice? Not printing "Parse results:" when we're in quiet mode. This way people could nearley-test to execute grammars in general. Would you like to implement this? Should be like a 2-line change.

sandiegoscott:
Do you mean output Parse Table only? So as not to lose the current addition of Parse Results only, maybe a single option to set what’s included—both (default), Table only, Results only? A bit cumbersome, perhaps, but anything that includes Parse Results only is okay with me.


On Oct 2, 2016, at 3:14 PM, Hardmath123 <notifications@github.com> wrote:

Actually, you know what would be nice? Not printing "Parse results:" when we're in quiet mode. This way people could nearley-test to execute grammars in general. Would you like to implement this? Should be like a 2-line change.

Hardmath123:
No, I mean, when in quiet mode it shouldn't print the heading "Parse results:" before the actual output.

sandiegoscott:
Pushed. As a test I redirected the output on the command line to file. In it are lots of escapes:

`'X', null`   becomes   `[32m'X'[39m, [1mnull[22m`

The reason?

Hardmath123:
The escapes are ANSI escape codes that colorize the terminal output. Good catch. There'll be a `require('util').inspect` somewhere. Ideally, it would have `color: false` when quiet and `color: true` when printing the parse table.

Also, you'll have to open a fresh PR.

PR #133, opened 2016-10-04
fixing undefined `this`

uriva:
function did not bind to enclosing scope but still used `this` which was undefined


Hardmath123:
Thanks.

Issue #135, opened 2016-10-15
optimizing - `for` faster than `forEach`??


ayapi:
Hi! thank you for nearley
maybe my english is very bad but please read

grammar that i'm trying to,  it has many many rules, very complex,
it can finish to parse without ambiguous, but it takes 45 seconds to parse 6000 tokens
(w/ tokenizer and custom-token).

but when i rewrote nearley source code `State.prototype.process`,
`forEach` to `for` and remove `me` closure variable,
like this,

before
```js
var me = this;
rules.forEach(function(r) {
```

after
```js
for(var i = 0; i < rules.length; i++) {
    var r = rules[i];
```

and use `this` instead of `me` in `for` block,
then, parsing time is 5 seconds.
45sec -> 5sec !?
it seems, parsing results(tree) are same..
i'm using nearley on Node.js v6.8.0.
i'm new to nearley, so i dont have confidence...
but if you feel up to it, please try / verify it. thank you.

Hardmath123:
Hmm, that sounds good to me! Can you give me your tokenizer and sample input so that I can try this out? Thanks!

ayapi:
tokenizer have any relation to this topic i think.
simply run below with `examples/javascript.ne` grammar,

```js
let i = 0;
let input = '';
while(i < 100) {
  input += `if (i > 0) {i = i - 1;}`
  i++;
}
let parser = new nearley.Parser(grammar.ParserRules, grammar.ParserStart);
parser.feed(input);
```

for -> 2sec
forEach -> 14sec

bobbybee:
@ayapi Is the patch #136 correct?

ayapi:
@bobbybee thank you! yes that's right!

Hardmath123:
Hey, @ayapi, can you update your nearley and try your examples again? Hopefully they should be faster.

ayapi:
i've confirmed on latest just now, it become faster than before. :ok_hand:
thank you, i'll close this issue

ayapi:
@Hardmath123 
i found this bottleneck using profiling features on node.js.
as you already know, `node --prof` and `node --prof-process` command.
https://nodejs.org/en/docs/guides/simple-profiling/
please make it useful for optimization

PR #136, opened 2016-10-16
Implement patch described in issue #135

bobbybee:
Tests pass and are somewhat faster than master; however, detailed performance evaluation was not done.


Hardmath123:
\o/

PR #145, opened 2016-12-06
Fix for strict mode

JakeSidSmith:
Line 73 causes the following error in strict mode: `In strict mode, function declarations cannot be nested inside a statement or block. They may only appear at the top level or directly inside a function body.`

This is simply fixed by assigning the function to a variable.

Would be good to get this fix in as we are running our entire app in strict mode, and since pulling in nearly, have lost some browser compatibility due to this error.

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -98,7 +98,7 @@ State.prototype.process = function(location, ind, table, rules, addedRules) {
                 } else {
                     return false;
                 }
-            }
+            };
 
             var leo = findLeo(this.reference, this.rule.name, this.data);
             if (leo) {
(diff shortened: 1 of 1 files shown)


Hardmath123:
Looks okay. Merged.

Hardmath123:
(Thanks.)

JakeSidSmith:
No problem. Thanks for the quick response. Can you tag me here when you release?

Hardmath123:
Already pushed to npm (looks like it hasn't updated on the website yet—caching issues?—but if you `npm install` you get 2.7.9, the latest).

JakeSidSmith:
Nice. Thanks very much. :)

PR #152, opened 2017-01-21
nearley-test: Provide `type: string` on --input option

### LINKED ISSUE #151: Single string matching (works with a latin character, but doesn't work with an integer)
```bash
$ echo 'test -> "a"' > ./grammar.ne
$ nearleyc ./grammar.ne --out ./grammar.js
$ nearley-test ./grammar.js --input 'a'
Table length: 2
Number of parses: 1
Parse Charts
Chart: 0
0: {test →  ● "a"}, from: 0

Chart: 1
0: {test → "a" ● }, from: 0


Parse results:
[ [ 'a' ] ]

```

```bash
$ echo 'test -> "1"' > ./grammar.ne
$ nearleyc ./grammar.ne --out ./grammar.js
$ nearley-test ./grammar.js --input '1'
Table length: 1
Number of parses: 0
Parse Charts
Chart: 0
0: {test →  ● "1"}, from: 0


Parse results:
[]

```

Whats the reason one produces a result and the other does not?

nanalan:
Resolves #151.

### DIFF
--- a/bin/nearley-test.js
+++ b/bin/nearley-test.js
@@ -19,6 +19,7 @@ var opts = nomnom
     .option('input', {
         abbr: 'i',
         help: "An input string to parse (if not provided then read from stdin)",
+        type: 'string',
     })
     .option('start', {
         abbr: 's',


nanalan:
For context, from the [nomnom docs](https://github.com/harthur/nomnom#type):
> If you don't want the option JSON-parsed, specify type "string".

Hardmath123:
Thanks.

PR #164, opened 2017-02-16
Remove experimental JavaScript parser

tjvr:
It's somewhat outdated and misleading, and seems to be confusing people.

The ES5 spec alone is very very big and scary; I'm not convinced that implementing a CFG for it is practical or worthwhile!


Hardmath123:
#sweetvibesofapproval

tjvr:
Yay xD

Issue #111, opened 2016-07-24
Very slow performance for JSON grammar.


bd82:
Hello.

### Background
I'm trying to add nearley to a benchmark of multiple javascript parsing libraries.
For this purpose I've implemented a JSON grammar using nearley
and have "fed" it my standard sample (±1,000 lines) json input.

### Issue
On my modern MacBook Pro (4ghz I7) parsing this sample takes ±6-7 seconds.
This is as much as **three order of magnitudes** slower than other Parsing libraries I've used.

[**Reproducing html page**](http://sap.github.io/chevrotain/performance/index_nearley.html
)
* This page prints the total parsing time to the **console**.
* It may take a few seconds until the page is responsive to allow opening the development tools.

### Sources:
* [Grammar](https://github.com/SAP/chevrotain/blob/47f40b1f464bae834dad1af5cbe3654671b5f73b/performance/jsonParsers/nearley/json_grammar.ne)
* [Generated Code](https://github.com/SAP/chevrotain/blob/47f40b1f464bae834dad1af5cbe3654671b5f73b/performance/jsonParsers/nearley/nearley_json_parser.js)
* [benchmark code](https://github.com/SAP/chevrotain/blob/47f40b1f464bae834dad1af5cbe3654671b5f73b/performance/index_nearley.html#L22)
* [1,000 lines sample input](https://github.com/SAP/chevrotain/blob/47f40b1f464bae834dad1af5cbe3654671b5f73b/performance/samples/1K_json_no_escaping.js)

### Extra Details
I'm even ignoring whitespace (pre-processing it out of the input) and literal string escaping handling to simplify the grammar. so its a strict subset of the JSON grammar which makes it **extremely** simple as the base JSON grammar is already quite simple...

### Questions
* Am I doing something incorrectly / missing something?
* Should this be examined using a separate tokenizer as well?

Thanks.

bobbybee:
What does nearley-test output? (Is it ambiguous?)

Hardmath123:
You should use a tokenizer, of course.

You have to realize that nearley is a table-based parser, which means it creates a fresh row in a large table for each token that comes in. This is the core of the Earley algorithm, which makes it so much more powerful than boring old naïve parsers.

[...]

PR #171, opened 2017-02-22
Make things Faster!

tjvr:
This is a PR to track some of the work I'm doing on rewriting nearley's internals :-)

### DIFF
--- a/bin/nearley-test.js
+++ b/bin/nearley-test.js
@@ -53,13 +53,12 @@ var writeTable = function (writeStream, parser) {
     writeStream.write("Table length: " + parser.table.length + "\n");
     writeStream.write("Number of parses: " + parser.results.length + "\n");
     writeStream.write("Parse Charts");
-    var chartNumber = 0;
     parser.table.forEach(
-        function (chart) {
+        function (column, chartNumber) {
+            console.log(column);
             writeStream.write("\nChart: " + chartNumber++ + "\n");
-            var stateNumber = 0;
-            chart.forEach(
-                function (state) {
+            column.states.forEach(
+                function (state, stateNumber) {
                     writeStream.write(stateNumber++ + ": " + state.toString() + "\n");
                 } )
         } )
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -46,22 +46,6 @@ State.prototype.nextState = function(data) {
     return state;
 };
 
-State.prototype.consumeTerminal = function(inp) {
-    var val = false;
-    if (this.rule.symbols[this.dot]) {                  // is there a symbol to test?
-       if (this.rule.symbols[this.dot].test) {          // is the symbol a regex?
-          if (this.rule.symbols[this.dot].test(inp)) {  // does the regex match
-             val = this.nextState(inp);  // nextState on a successful regex match
-          }
-       } else {   // not a regex, must be a literal
-          if (this.rule.symbols[this.dot].literal === inp) {
-             val = this.nextState(inp);  // nextState on a successful literal match
-          }
-       }
-    }
-    return val;
-};
-
 State.prototype.build = function() {
     var children = [];
     var node = this;
@@ -84,7 +68,8 @@ function Column(grammar, index) {
     this.grammar = grammar;
     this.index = index;
     this.states = [];
-    this.wants = {}; // states that want some token
+    this.wants = {}; // states indexed by the non-terminal they expect
+    this.scannable = []; // list of states that expect a token
     this.completed = {}; // states that are nullable
 }
 
@@ -116,10 +101,14 @@ Column.prototype.process = function(nextColumn) {
             }
 
         } else {
-            // predict
+            // queue scannable states
             var exp = state.rule.symbols[state.dot];
-            if (typeof exp !== 'string') { continue; }
+            if (typeof exp !== 'string') {
+                this.scannable.push(state);
+                continue;
+            }
 
+            // predict
             if (wants[exp]) {
                 wants[exp].push(state);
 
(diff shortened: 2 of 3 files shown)


tjvr:
So far this isn't as dramatically faster as I'd hoped for.

However, this might just be because I'm not looking at the right sort of benchmarks. Currently I'm using the test runs as a proxy for performance, which isn't terribly informative! Ideally I'd test this on a large tosh grammar with a tokenizer in front of it, perhaps. :)

bd82:
See a previous issue related to performance.
https://github.com/Hardmath123/nearley/issues/111

I've used (and linked) a few benchmark pages there using nearley & JSON Grammar
plus 1,000 lines JSON input.

You can also use the main benchmark pages which compares relative performance
of different JavaScript Parser Generators/Toolkits/Combinators.
http://sap.github.io/chevrotain/performance/

The UX there allows selecting only the parser toolkit you want.
And it also includes a link to the github repo so you can just clone it locally, modify the code (load different nearley.js file) and run it by opening the performance/index.html file.

bd82:
For advanced optimizations also try
IRHydra
http://mrale.ph/irhydra/2/

It is a tool which helps you figure out de-optimization reasons under V8 engine.
It was very useful when I've performed performance optimizations in the past.

tjvr:
@bd82 No thanks. JSON is not a useful benchmark.

bd82:
> JSON is not a useful benchmark.

Why not?

Obviously it does not cover the full scope of Nearley capabilities as it is a very simple
grammar which only requires LL(1) and does not have ambiguities or left recursion.
So it can only be **part of** a benchmark suite not the whole suite.

But it has alternation, repetition and production nesting. These rules are probably enough
to describe the Python programing language which is also LL(1) as it is a [design goal](https://www.python.org/dev/peps/pep-3099/) of python.

It is also simple to implement so you can easily compare Nearley performance to other parsing toolkits or even compare to hand built JSON parsers.

If you can get a simple grammar to be faster won't that influence the more complex ones too?

**What would you define as a useful benchmark?**

tjvr:
@bd82 This is not the right place for this conversation, sorry.

tjvr:
Actually, this is a pretty compelling set of improvements:
```sh
$ git checkout master
$ node test/benchmark.js 
  ✔ calculator arithmetic2 x 776 ops/sec ±2.45% (84 runs sampled)
  ✔ json test2.json x 325 ops/sec ±0.36% (88 runs sampled)
  ✔ tosh ex1 x 37.43 ops/sec ±0.56% (63 runs sampled)
$ git checkout opt
$ node test/benchmark.js 
  ✔ calculator arithmetic2 x 1,910 ops/sec ±4.61% (84 runs sampled)
  ✔ json test2.json x 705 ops/sec ±0.55% (86 runs sampled)
  ✔ tosh ex1 x 342 ops/sec ±1.46% (84 runs sampled)
Done in 17.40s.
```
* The calculator benchmark is ~2x faster
* The json benchmark is ~2x faster
* The tosh benchmark is nearly an order of magnitude faster! (This PR focuses on large grammars, so this is not surprising.)

Bear in mind these are only small examples. :)

tjvr:
However, I don't think this is quite mergeable yet; I think it may still have some bugs we need to track down...

tjvr:
I'm a lot more confident about this now I've tracked down #180. (Although I'm still a little confused as to why this PR seems to cause nearley to use more memory!)

tjvr:
I added a commit 85ca34b that GCs unused states :D
Since State now keeps track of the states that predicted it -- the `wantedBy`
attribute -- states which are or may be part of the final parse are always
kept.

When we support rewind(), we'll have to make this optional (based on a `{ rewindable: true }` option given to Parser?)

You can see the difference if you run `node test/profile.js`.

tjvr:
I think this is good now @Hardmath123 -- want to take a look at some point? :)

bd82:
Tested this on a 45K (1,000) lines JSON.
Using the lexerless JSON grammar example (minus the grammar actions).

with 2.7.14 I got around 1.3 ops/second.
with 2.7.14 + these optimizations I got 4.71 ops/second.

so very very nice 👍

tjvr:
@bd82 Thanks for testing :-)

I'd be very interested to see how this affects your JSON plus lexer parser. :-)

bd82:
I will try to test tomorrow.

bd82:
With Tokenizer:

About **40 Ops**/second with **both** 2.7.13 and 2.7.14
With 2.7.14 + these optimizations: about **95 Ops**/second.

👍

bd82:
### Suggestion for possible optimization:
The [State Object](https://github.com/Hardmath123/nearley/blob/3b415764680df370317f1de3519bf45263a3b5fa/lib/nearley.js#L35) is initialized using the **"new"** keyword.

I'm not sure if this will help, but I have seen improvements in similar scenarios when
using more basic structures, I.E:
```
var state = {rule:X, expect:Y, reference:Z, data:[]}
```
and instead of putting the methods on the State Prototype:
```Javascript
State.prototype.consumeNonTerminal = function(inp) {...}
```
Using standard functions which accept the state object:
```Javascript
function consumeNonTerminal(inp, state) {...}
```
I tried doing it myself but messed up the refactoring and ended up with an infinite loop.
I may find time later this week to try again. But you may have better luck trying this
(assuming that hopefully it would help).

tjvr:
@bd82 Thans for benchmarking with the tokenizer! (I tried it myself but couldn't get it to work :-) )

tjvr:
I decided this still wasn't fast enough, so I bring you 7fc11d4 :
```
  ✔ calculator arithmetic2 x 3,918 ops/sec ±1.97% (81 runs sampled)
  ✔ json test2.json x 1,636 ops/sec ±1.53% (86 runs sampled)
  ✔ tosh ex1 x 448 ops/sec ±1.29% (86 runs sampled)
```

Hardmath123:
:D

tjvr:
:D :D

cmcaine:
I get slightly better performance with chevrotain's benchmark on e96a532 than on v2.11 (~68ops/s vs ~55 ops/s)

I don't get anywhere near 1000 ops/s. Presumably you and bd82 are measuring different ops? Maybe because bd82's json is so huge?

My mirror of the performance page (uses v2.11): https://cmcaine.github.io/chevrotain/performance/

bd82:
Yes @cmcaine.

There is not much point in comparing **absolute** ops/second over **different** machines and **different** inputs.

It is the **relative** performance numbers that are actually meaningful.

cmcaine:
Yes, of course :)

I wanted to highlight a possible performance regression in nearley 2.11 compared to this PR.

tjvr:
> a possible performance regression

If you think you've found one -- our preferred benchmarks can be run using `npm run benchmark`, btw -- then please open a new issue. :-)

cmcaine:
The benchmarks are dissimilar and I can't compile the new version of json.ne at this commit:

```
in nearley on e96a532...
› git tag faster

in nearley on tags/faster
› npm install && npm install moo

in nearley on tags/faster
› git checkout master examples/json.ne

in nearley on tags/faster
› ./bin/nearleyc.js examples/json.ne | node
[stdin]:78
    {"name": "number", "symbols": [number], "postprocess": function(d) { return parseFloat(d[0].value) }},
                                   ^

ReferenceError: number is not defined
```

Running sample1k.json through the old json.ne gives ~9 op/s, but there's no lexer on that.

PR #175, opened 2017-02-24
Benchmark suite

tjvr:
Today's patch: adding a lil benchmark suite.

Yes, benchmarking nearley's JSON example against a native JSON parser is
hilarious and stupid. But it demonstrates that the benchmarks work (-:

I've also refactored the existing tests to invoke nearleyc directly, rather than through a subprocess. This makes `npm test` a fair bit faster (4s -> 2s over here).

@hardmath123 What do you think? :-)

---

    $ npm run benchmark

### DIFF
--- a/test/_shared.js
+++ b/test/_shared.js
@@ -1,0 +1,49 @@
+
+var fs = require('fs');
+
+var nearley = require('../lib/nearley.js');
+var Compile = require('../lib/compile.js');
+var parserGrammar = require('../lib/nearley-language-bootstrapped.js');
+var generate = require('../lib/generate.js');
+
+function parse(grammar, input) {
+    if (grammar.should) {
+        grammar.should.have.keys(['ParserRules', 'ParserStart']);
+    }
+    var p = new nearley.Parser(grammar.ParserRules, grammar.ParserStart);
+    p.feed(input);
+    return p.results;
+}
+
+function nearleyc(source) {
+    // parse
+    var results = parse(parserGrammar, source);
+
+    // compile
+    var c = Compile(results[0], {});
+
+    // generate
+    var compiledGrammar = generate(c, 'grammar');
+
+    // eval
+    return evalGrammar(compiledGrammar);
+}
+
+function evalGrammar(compiledGrammar) {
+    var f = new Function('module', compiledGrammar);
+    var m = {exports: {}};
+    f(m);
+    return m.exports;
+}
+
+function read(filename) {
+    return fs.readFileSync(filename, 'utf-8');
+}
+
+module.exports = {
+    read: read,
+    nearleyc: nearleyc,
+    parse: parse,
+    evalGrammar: evalGrammar,
+};
+
--- a/test/_shared.js
+++ b/test/_shared.js
@@ -15,7 +15,7 @@ function parse(grammar, input) {
     return p.results;
 }
 
-function nearleyc(source) {
+function compile(source) {
     // parse
     var results = parse(parserGrammar, source);
 
@@ -42,7 +42,7 @@ function read(filename) {
 
 module.exports = {
     read: read,
-    nearleyc: nearleyc,
+    compile: compile,
     parse: parse,
     evalGrammar: evalGrammar,
 };
--- a/test/benchmark.js
+++ b/test/benchmark.js
@@ -115,6 +115,8 @@ suite.on('cycle', function(event) {
 
   if (bench.error) {
     console.log(colors.red("✘"), bench.name);
+    console.log(colors.red(bench.error.stack));
+    console.log('');
   } else {
     var opsPerSec = formatNumber(hz.toFixed(hz < 100 ? 2 : 0)) + ' ops/sec ' + pm + stats.rme.toFixed(2) + '%';
     console.log(colors.green("✔"), padName(bench.name), colors.blue(opsPerSec));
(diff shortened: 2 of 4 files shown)


test/_shared.js
```diff
@@ -0,0 +1,49 @@
+
+var fs = require('fs');
+
+var nearley = require('../lib/nearley.js');
+var Compile = require('../lib/compile.js');
+var parserGrammar = require('../lib/nearley-language-bootstrapped.js');
+var generate = require('../lib/generate.js');
+
+function parse(grammar, input) {
+    if (grammar.should) {
```
Hardmath123: Is this really necessary? :P

test/benchmark.js
```diff
@@ -0,0 +1,110 @@
+
+var fs = require('fs');
+var path = require('path');
+var child_process = require('child_process')
```
Hardmath123: As far as I can tell, this isn't actually used in this file? Unless I missed something.

test/benchmark.js
```diff
@@ -0,0 +1,110 @@
...
+
+var Benchmark = require('benchmark');
+var colors = require('colors/safe');
```
Hardmath123: (See below, same for `formatNumber`.)

test/benchmark.js
```diff
@@ -0,0 +1,110 @@
...
+  , read = shared.read;
+
+
+// Define benchmarks
+
+function addTest(parserName, parser, exampleInputs) {
+    exampleInputs.forEach(function(inputPath) {
+        var input;
+        var exampleName;
+        if (/^test\//.test(inputPath)) {
```
Hardmath123: Can't you just do `read("test1.json")` in the test specification? This seems like a hack that will cause trouble in the future.

test/benchmark.js
```diff
@@ -0,0 +1,110 @@
...
+addTest('native JSON.parse', JSON.parse, [
+   'test/test1.json',
+   'test/test2.json',
+])
+*/
+
+
+// Run & report results
+
+var longestName = Math.max.apply(null, suite.map(function(bench) {
```
Hardmath123: (See below for these two functions.)

test/benchmark.js
```diff
@@ -0,0 +1,110 @@
...
+    var stats = bench.stats;
+    var hz = bench.hz; // Hz -- ops per sec
+    var pm = '\xb1';
+
+    if (bench.error) {
+        console.log(colors.red("✘"), bench.name);
+        console.log(colors.red(bench.error.stack));
+        console.log('');
+    } else {
+        var opsPerSec = formatNumber(hz.toFixed(hz < 100 ? 2 : 0)) + ' ops/sec ' + pm + stats.rme.toFixed(2) + '%';
```
Hardmath123: (This is below.)

Can't we just use benchmark.js's built-in functionality, with `console.log(String(event.target))`? I'm scared that we're defining like three functions and importing a whole module just to format a line of output. This seems to work fine:

```
    var bench = event.target;
    if (bench.error) {
        console.log(bench.name);
        console.log(bench.error.stack);
        console.log('');
    } else {
        console.log(String(event.target))
    }
```

Hardmath123:
Looks good! I'm on board for the changes to `launch.js`, including the refactoring that happened with `_shared.js`. For `benchmark.js`, I added a few comments for things I was confused about. Let me know what you think (they're all small things).

test/benchmark.js
tjvr: You're probably right! Thanks :)

test/benchmark.js
tjvr: Yes, but then I can't infer the name of the test. :P I should probably have another look at this; any suggestions?

test/benchmark.js
tjvr: We could, but then the colors aren't pretty, and the names aren't aligned! :P

I find this significantly easier to read, and I picked `colors` because mocha is using it anyway. (Although sadly a different version.) Benchmark's API here isn't going to change, so this shouldn't be a bother anytime soon. :)

test/_shared.js
tjvr: Hey, it was in launch.js. I just copied it over :P

Hardmath123: I didn't actually write launch.js, so that's totally not on me! :P I think we should take it out, though.

test/benchmark.js
Hardmath123: I think clipping long strings with something like `JSON.stringify(input.substring(0, 10) + '...')` should be fine?

test/benchmark.js
Hardmath123: Hmm. I'm not a big fan of importing random tiny modules just to make things marginally nicer (nearley currently has three dependencies, all of which are "can't-live-without" code).

I guess my opinion is that `String(event.target)` is good enough and built-in (and how often will you be reading those, anyway?). But whatever you want to do. :-)

test/benchmark.js
jupiterio: _Pssst #135_

Hardmath123: …this `forEach` is really not likely to be a bottleneck… :P

jupiterio: Oh, ok, I didn't really know 📦

tjvr:
Okay @hardmath123, I addressed all your comments. Happy? :)

Hardmath123:
> Happy? :)

Absolutely! Thanks Tim.

PR #180, opened 2017-02-27
Left-recursive EBNF expansion

tjvr:
When testing #171 I found I was running out of memory when parsing a large JSON sample (~1kB), using the example scannerless JSON parser in the repo.

I realised that we're expanding EBNF rules like
```
string -> "\"" validChar:* "\""
```
into
```
string$ebnf$1 -> 
               | validChar string$ebnf$1
```

Since Earley loves left-recursion, I think it would be better to expand these left-recursively instead! In particular, I *think* this avoids some weird n^2 behaviour when parsing long strings.

Anyway, even on master this patch appears to speed up a 1k JSON sample by ~3x, so this seems worthwhile :-)

@hardmath123 What do you think?

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -200,8 +200,8 @@ function Compile(structure, opts) {
             [{
                 tokens: [token.ebnf],
             }, {
-                tokens: [token.ebnf, name],
-                postprocess: {builtin: "arrconcat"}
+                tokens: [name, token.ebnf],
+                postprocess: {builtin: "arrpush"}
             }],
             env
         );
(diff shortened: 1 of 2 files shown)


Hardmath123:
Yes! This is a thing I've been meaning to do for a while. Thank you!

Hardmath123:
(I published v2.7.14 on npm because I think this is really worthwhile. Also, I recompiled the bootstrapped grammar so now compiling is faster as well.)

tjvr:
> (I published v2.7.14 on npm because I think this is really worthwhile.

:D

bd82:
Awesome 👍 

I can confirm a 300% (X4) performance increase on a 45k input JSON file 
between 2.7.13 and 2.7.14.

tjvr:
@bd82 Neat! What do you get with #171?

bd82:
@tjvr I will answer on #171

PR #182, opened 2017-02-28
Added parentheses.ne

vietlq:
Added grammar to parse matching parentheses, brackets, square brackets and pairs of "<" ">".

### DIFF
--- a/test/launch.js
+++ b/test/launch.js
@@ -116,17 +116,15 @@ describe("nearleyc", function() {
             parse(parentheses, passCases[i]).should.deep.equal([true]);
         }
 
-        /*
         var failCases = [
             ' ',
             '[}',
             '[(){}><]',
-            '((((())))(())()'
+            '(((())))(()))'
         ];
 
         for (let i in failCases) {
             (function() { parse(parentheses, failCases[i]); }).should.throw(Error);
         }
-        */
     });
 });
--- a/test/launch.js
+++ b/test/launch.js
@@ -105,9 +105,28 @@ describe("nearleyc", function() {
     it('parentheses', function() {
         // Try compiling the grammar
         var parentheses = compile(read("examples/parentheses.ne"));
-        var parenthesesTest1 = '[(((<>)()({})())(()())(())[])]';
-        var parenthesesTest2 = '<<[([])]>([(<>[]{}{}<>())[{}[][]{}{}[]<>[]{}<>{}<>[]<>{}()][[][][]()()()]({})<[]>{(){}()<>}(<>[])]())({})>';
-        parse(parentheses, parenthesesTest1).should.deep.equal([true]);
-        parse(parentheses, parenthesesTest2).should.deep.equal([true]);
+        var passCases = [
+            '()',
+            '[(){}<>]',
+            '[(((<>)()({})())(()())(())[])]',
+            '<<[([])]>([(<>[]{}{}<>())[{}[][]{}{}[]<>[]{}<>{}<>[]<>{}()][[][][]()()()]({})<[]>{(){}()<>}(<>[])]())({})>'
+        ];
+
+        for (let i in passCases) {
+            parse(parentheses, passCases[i]).should.deep.equal([true]);
+        }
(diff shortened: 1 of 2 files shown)


test/launch.js
```diff
@@ -103,4 +101,13 @@ describe("nearleyc", function() {
...
         parse(classicCrontab, crontabTest).should.deep.equal([JSON.parse(crontabResults)]);
     });
+
+    it('parentheses', function() {
+        // Try compiling the grammar
+        var parentheses = compile(read("examples/parentheses.ne"));
+        var parenthesesTest1 = '[(((<>)()({})())(()())(())[])]';
+        var parenthesesTest2 = '<<[([])]>([(<>[]{}{}<>())[{}[][]{}{}[]<>[]{}<>{}<>[]<>{}()][[][][]()()()]({})<[]>{(){}()<>}(<>[])]())({})>';
+        parse(parentheses, parenthesesTest1).should.deep.equal([true]);
+        parse(parentheses, parenthesesTest2).should.deep.equal([true]);
```
tjvr: Is it possible to make this test match something more interesting? We want to make sure here that nearley is working as expected. 

(A valid answer here would be "we're just testing the recogniser here and not the parser". I would be fine with that. :) )

tjvr:
Nice, thanks! :)

test/launch.js
vietlq: Ok, will add more test cases and also failed cases too.

vietlq: Please kindly help me with Chai:

```

        var failCases = [
            ' ',
            '[}',
            '[(){}><]',
            '((((())))(())()'
        ];

        for (let i in failCases) {
            parse(parentheses, failCases[i]);
            (function() { parse(parentheses, failCases[i]); }).should.throw(Error);
        }
```

I will get this error:

```
  1) nearleyc parentheses:
     Error: nearley: No possible parsings (@0: ' ').
      at Parser.feed (lib/nearley.js:273:23)
      at parse (test/_shared.js:14:7)
      at Context.<anonymous> (test/launch.js:126:13)
```

If I remove the line `parse(parentheses, failCases[i]);`, I will get:

```
  1) nearleyc parentheses:
     AssertionError: expected [Function] to throw Error
      at Context.<anonymous> (test/launch.js:127:76)
```

That's really confusing. Clearly an instance of Error is being thrown. Thanks.

vietlq: https://github.com/chaijs/chai/issues/559:

```
But basically this is related to the behavior when using an Error for the check. When doing this, Chai does a strict (===) comparison to check if both instances are the same.
```

vietlq:
@tjvr @Hardmath123 if you uncomment `failCases`, you will uncover the ugly side of `Chai`.

vietlq:
Please don't accept yet. I found a bug:

```
parse(parentheses, '((((())))(())()');
```

There's an unbalanced "(", the 2nd last char.

vietlq:
Why do we have different behaviour by the parser?

```
$ nearley-test -i "[}" parentheses.js
/usr/local/lib/node_modules/nearley/lib/nearley.js:278
            throw err;
            ^

Error: nearley: No possible parsings (@1: '}').
```

These will have no issue, even they are invalid:

```
$ nearley-test -i "" parentheses.js
$ nearley-test -i "((((())))(())()" parentheses.js
```

Hardmath123:
In the latter example, there is the possibility that by adding more characters, you could get a valid string. The solution is to check whether `parser.results` is empty or not. The error is only thrown when you add an invalid character, that is, when _no matter what_ you will never get a valid string by adding more characters.

vietlq:
@Hardmath123 I can see from command line test that `parser.results === []` and not `parser.results === [true]`. So when we do parsing, that should be expected behaviour? You are right regarding possibility to add ")" to the end and solve the issue.

Hardmath123:
Yes, exactly.

vietlq:
@Hardmath123 Thanks. I updated the test cases accordingly and also added comments for clarity.

PR #184, opened 2017-03-01
Use UMD

### LINKED ISSUE #179: Don't check for `window`
See <https://github.com/Hardmath123/nearley/blob/master/lib/compile.js#L2>:
```js
if (typeof(window) === 'undefined') {
    var nearley = require('../lib/nearley.js');
} else {
    var nearley = window.nearley;
}
```
Some testing frameworks define `window`, so this check fails.

I think there are better conditions we could use here, e.g. the `typeof module !== 'undefined' && typeof module.exports !== 'undefined'` check seen at the bottom of the same file. :)

tjvr:
Fixes #179.

(I've disabled AMD for now, because we don't know of anyone who's using it; I don't want to claim to support stuff that we haven't tested!)

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -1,4 +1,13 @@
-(function () {
+(function(root, factory) {
+    // if (typeof define === 'function' && define.amd) {
+    //     define([], factory);
+    if (typeof module === 'object' && module.exports) {
+        module.exports = factory();
+    } else {
+        root.nearley = factory();
+    }
+}(this, function() {
+
 function Rule(name, symbols, postprocess) {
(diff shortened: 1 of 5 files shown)


lib/nearley.js
```diff
@@ -1,4 +1,13 @@
-(function () {
+(function(root, factory) {
+    // if (typeof define === 'function' && define.amd) {
```
vietlq: I wouldn't recommend keeping comments like this.

tjvr: If you want to go ahead and test nearley under AMD, go right ahead :)

Issue #183, opened 2017-03-01
Make GC thing optional


Hardmath123:
It should be disabled when doing nearley-test, or (eventually) when rewinding.

tjvr:
When `rewindable` is enabled, that is. Hmm; should these be options given to parser?

Hardmath123:
Yeah, you should have to specify options when constructing a `new Parser()`. That way nearley-test can do the right thing.

tjvr:
Any suggestion for what this should be called?

wizzard0:
something like `preserveState` maybe...

tjvr:
This is now merged. :)

PR #185, opened 2017-03-01
Add history option & refactor Grammar construction

tjvr:
First, we allow Parser to accept a Grammar directly. This is the preferred way, since it avoids constructing the Grammar every time we want to parse something (which has a cost O(n) in the size of the grammar). I've modified the benchmarks to cache the Grammar object, so there should be a minor speed bump there (should be more for larger grammars I guess?).

To make the tests easier to write, I added `Grammar.fromCompiled()`. There are thus three ways to get a grammar:

    Grammar(rules, start) // note: list of Rule objects
    Grammar.fromCompiled({ParserRules,ParserStart})
    Grammar.fromCompiled(ParserRules, ParserStart)

This is also used for backwards-compatibility in the Parser constructor, so you can still construct a Parser the old way (even though caching your Grammar is preferred).

Then, we add a lightweight options dict to Parser, containing a sole `history` option. If it is true, we include all the columns in the Earley table, disabling the memory optimisation that removes old unusued states.

So parser can now be initialised four ways:

    Parser(Grammar)
    Parser(Grammar, options)
    Parser(ParserRules, ParserStart)
    Parser(ParserRules, ParserStart, options)

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,11 +290,13 @@ Using a parser
 
 nearley exposes the following API:
 
-    var grammar = require("generated-code.js");
+    var generated = require("generated-code.js");
     var nearley = require("nearley");
 
+    var grammar = nearley.Grammar.fromCompiled(generated)
+
     // Create a Parser object from our grammar.
-    var p = new nearley.Parser(grammar.ParserRules, grammar.ParserStart);
+    var p = new nearley.Parser(grammar);
 
     // Parse something
     p.feed("1+1");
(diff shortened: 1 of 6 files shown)


tjvr:
This is for #183.

tjvr:
If we do a v3 with a new API, I'd like to force giving Parser a Grammar object. But that's a story for another time... :-) #186

Hardmath123:
Can the option be called `keepHistory` instead? It sounds more readable. Also, could you slip a sentence into the appropriate section of `README.md`, documenting how and when to use `keepHistory`? Thanks.

tjvr:
> the appropriate section of README.md

Sure! Whereabouts would you like it to go? :)

Hardmath123:
In this section, I think --- https://github.com/Hardmath123/nearley#using-a-parser

tjvr:
Updated. :)

Hardmath123:
If you don't mind, I rewrote that section of the README a bit. Otherwise, everything has been merged. :-)

tjvr:
> I rewrote that section of the README a bit.

Great! I did *not* get your voice :P

vietlq:
Guys, today's parser runs faster!

tjvr:
@vietlq Noticeably faster? I'm glad to hear it :D #171 must have been worth it then!

PR #187, opened 2017-03-01
Added grammar for COOL recursive comments

vietlq:
Grammar for parsing content of recursive comments of COOL (The Classroom Object-Oriented Language)

* Treat everything inside the outmost (**) as content
* Empty comment: `(**)`
* Recursive comments: `(*(**)*)` => `(**)`, `(*(*a*)(*b*)*)` => `(*a*)(*b*)`
* Invalid comments: `(*)*)`, `(*(*)`, `(*(**)`, `(*(*)*)`
* For details, check: http://theory.stanford.edu/~aiken/software/cool/cool.html

### DIFF
--- a/examples/cool_comments.ne
+++ b/examples/cool_comments.ne
@@ -1,0 +1,23 @@
+# Grammar for parsing content of recursive comments of COOL (The Classroom Object-Oriented Language)
+# Treat everything inside the outmost (**) as content
+# Empty comment: (**)
+# Recursive comments: (*(**)*) => (**), (*(*a*)(*b*)*) => (*a*)(*b*)
+# Invalid comments: (*)*), (*(*), (*(**), (*(*)*)
+# For details, check: http://theory.stanford.edu/~aiken/software/cool/cool.html
+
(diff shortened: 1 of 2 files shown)


Hardmath123:
I'm happy to merge the grammar, but I'm iffy about merging the test.

How much value do you think this additional test adds? I'm all for more tests, but I feel they should test orthogonal features rather than re-testing features that are already tested. (cc @tjvr)

Also, @vietlq, I feel we should prefer example parsers for well-known languages over academic examples. This is partly because it is easier for people to understand examples of languages they already know, and partly because an example parser for a real-world language might someday be useful for someone! :-)

vietlq:
Then I should rename to: OCaml nested comments

https://ocaml.org/learn/tutorials/basics.html#Comments

Also add some files to test the code. Sounds good? Not academic anymore.

vietlq:
I would appreciate an example how to improve the tests

vietlq:
I will reserve this for a newbie's tutorial.

PR #191, opened 2017-03-07
Add rewind()

tjvr:
First off, we require calling finish() before reading `results`. This is a deliberate **API change**.
Currently, we "incrementally keep track of results". This means feeding the input in multiple chunks has a performance penalty.
I've patched our Stream implementation, so users of that API won't even notice this change!
Everyone else will have to drop in a `p.finish()` line before they read `p.results`.

Then, I added rewind(index), which requires the `keepHistory` option. It resets the parser state to the state after seeing that token. So if you feed it 10 tokens, rewind(10) has no effect. [rewind(11) is an error; you can't rewind forwards!]

Note that this is *not a reset* method. If you want to reset a parser, throw it away and initialise a new one. Assuming you're using `Grammar` directly this is basically free. If you're using an Earley parser, you shouldn't care about allocating one single object. :-)

### DIFF
--- a/examples/tosh.ne
+++ b/examples/tosh.ne
@@ -386,7 +386,7 @@ block -> "else" {% d => ["else"] %}
 _ -> [ ]:* {% d => null %}
 __ -> [ ]:+ {% d => null %}
 
-string -> "'hello'"
+string -> "'hello'"             {% d => 'hello' %}
 number -> digits                {% d => parseInt(d[0]) %}
 number -> digits [.] digits     {% d => parseFloat(d.join('')) %}
 
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -267,11 +267,32 @@ Parser.prototype.feed = function(chunk) {
     }
 
     this.current += chunkPos;
+
     // Incrementally keep track of results
     this.results = this.finish();
 
     // Allow chaining, for whatever it's worth
     return this;
+};
+
+Parser.prototype.rewind = function(index) {
+    if (!this.options.keepHistory) {
+        throw new Error('set option `keepHistory` to enable rewinding')
+    }
+    if (this.current < this.index) {
+        // TODO: api -- consider silently succeeding?
+        throw new Error('cannot rewind forward!')
+    }
+    /*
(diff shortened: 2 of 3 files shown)


tjvr:
@Hardmath123 thoughts? As an API change this might have to wait for v3... :/

Hardmath123:
Hmm… I don't want to break things by requiring `.finish()`. Why is that a prerequisite for having `.rewind()`? Can't we just run `.finish()` at the end of `.rewind()` to fix results? Also, what are the possible uses for `.rewind()` besides syntax-highlighting tricks?

(Sorry that ended up being a log of questions! :P)

tjvr:
> Why is that a prerequisite for having .rewind()?

I suppose it's actually not, but I want to do that also, and I thought I might as well do it first :P I'll drop that commit from this PR, and instead add it to my stack of API changes...

I can't think of other uses off the top of my head. But in general being to re-parse only some of the input seems useful. We could perhaps instead have a CachingParser, which re-uses the beginning of the previous parse where possible; and let rewind() remain only as part of the private API. But I'd rather provide rewind(), and let people implement their own logic on top of that if they so choose.

tjvr:
Updated. :)

Hardmath123:
Sounds good to me. Thanks, Tim!

Hardmath123:
```
nearleyc
  ✓ should build test parser (check integrity) (122ms)
  ✓ should build for CoffeeScript (260ms)
  ✓ calculator example
  ✓ csscolor example (47ms)
  ✓ exponential whitespace bug
  ✓ nullable whitespace bug
  ✓ percent bug
  ✓ tokens
  ✓ leo bug
  ✓ json example compiles
  ✓ json test1
  ✓ json test2
  ✓ tosh example (174ms)
  ✓ classic crontab
  ✓ parentheses
Parser
  ✓ can rewind
  ✓ won't rewind without `keepHistory` option
```

I feel we might want to think about what the headings should be. The line between "nearleyc" and "Parser" seems kind of arbitrary.

tjvr:
At some point, probably. I figured a different section was justified since these are testing the API, not the compiler or parsing algorithm. :-)

PR #197, opened 2017-03-10
Fix: Allow SystemJS to correctly infer module type

### LINKED ISSUE #196: SystemJS fails to correctly detect module type
SystemJS fails to detect the type of the `nearley` module, assuming that it's AMD. It then tries to run it (incorrectly) as AMD, causing module loading to fail.

This is because of #184: Instead of removing the old AMD header in `lib/nearley.js` it simply comments it out. SystemJS detects module type with a regex- it isn't smart enough to ignore comments and as such, due to the [module detection order](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#module-format-detection), it gets incorrectly recognized as AMD.

I confirmed this by removing that commented out code, and `nearley` was successfully detected as CommonJS.

In SystemJS this can be mitigated with explicitly defining the module type in the `meta` config, but we'd prefer if the automatic detection was working because using meta is a bit of a hack.

quigleyj97:
This PR resolves #196 
The comments I've removed were fooling SystemJS into thinking that nearley was an AMD module.

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -1,6 +1,4 @@
 (function(root, factory) {
-    // if (typeof define === 'function' && define.amd) {
-    //     define(['nearley'], factory);
     if (typeof module === 'object' && module.exports) {
         module.exports = factory(require('./nearley'));
     } else {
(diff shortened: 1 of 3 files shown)


tjvr:
Looks good; all the tests still pass.

Thanks! :)

tjvr:
Should anyone desire AMD support in the future; send us a PR :)

PR #198, opened 2017-03-10
Make token matching semantics user-definable

tjvr:
As per our conversation @hardmath123.

This makes the custom tokens feature a little more useful; you can still define your own tokens with the %-syntax just as before, but now you have complete control over how symbols get matched with tokens inside the parser.

I couldn't detect any performance difference using our benchmarks.

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -177,6 +177,12 @@ Grammar.fromCompiled = function(rules, start) {
 }
 
 
+function defaultMatcher(expect, token) {
+    // either regex or literal
+    return expect.test ? expect.test(token) : expect.literal === token
+}
+
+
 function Parser(rules, start, options) {
     if (rules instanceof Grammar) {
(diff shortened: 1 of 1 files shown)


Hardmath123:
I'm wondering whether it should be named "matcher" or "match". I think "match" is better.

tjvr:
I'll call it whatever you want :-)

Hardmath123:
Well, I was asking you! :P I feel functions tend to be named verbs — "match" — but of course the key in the object is kind of a noun, since it is telling you what the match*er* is. Naming this right is important because it's user-facing so we can't change it later.

tjvr:
Yes, it should be a verb. I'd be more concerned about whether "match" is the right word in the first place :-)

Hardmath123:
`test` is both a noun and a verb… :P

tjvr:
Closing in favour of #220.

PR #201, opened 2017-03-11
Adds option to pass token end locations

EricMCornelius:
Hi there,

I've been working on an expression parsing language with syntax highlighting and thus found a need for the end locations for each rule application to be passed to the postprocessor functions.  Not sure if there's a better alternative here, but it seemed sensible to thread the parser options reference down to the Column and State objects for future configuration flexibility.

Not sure if this is the best approach, as I've just started familiarizing myself with this project yesterday, so please let me know if there is a preferable alternative way to achieve this.

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -30,12 +30,13 @@ Rule.prototype.toString = function(withCursorAt) {
 
 
 // a State is a rule at a position from a given starting point in the input stream (reference)
-function State(rule, dot, reference, wantedBy) {
+function State(rule, dot, reference, wantedBy, options) {
     this.rule = rule;
     this.dot = dot;
     this.reference = reference;
     this.data = [];
     this.wantedBy = wantedBy;
+    this.options = options;
     this.isComplete = this.dot === rule.symbols.length;
 }
 
@@ -44,7 +45,7 @@ State.prototype.toString = function() {
 };
 
 State.prototype.nextState = function(data) {
-    var state = new State(this.rule, this.dot + 1, this.reference, this.wantedBy);
+    var state = new State(this.rule, this.dot + 1, this.reference, this.wantedBy, this.options);
     state.left = this;
     state.right = data;
     if (state.isComplete) {
@@ -64,16 +65,22 @@ State.prototype.build = function() {
     return children;
 };
 
-State.prototype.finish = function() {
+State.prototype.finish = function(pos) {
     if (this.rule.postprocess) {
-        this.data = this.rule.postprocess(this.data, this.reference, Parser.fail);
+        if (this.options.fullLocation) {
+            this.data = this.rule.postprocess(this.data, [this.reference, pos + 1], Parser.fail);
+        }
+        else {
+            this.data = this.rule.postprocess(this.data, this.reference, Parser.fail);
+        }
     }
 };
 
 
-function Column(grammar, index) {
+function Column(grammar, index, options) {
     this.grammar = grammar;
     this.index = index;
+    this.options = options;
     this.states = [];
     this.wants = {}; // states indexed by the non-terminal they expect
     this.scannable = []; // list of states that expect a token
@@ -81,7 +88,7 @@ function Column(grammar, index) {
 }
 
 
-Column.prototype.process = function(nextColumn) {
+Column.prototype.process = function(pos) {
     var states = this.states;
     var wants = this.wants;
     var completed = this.completed;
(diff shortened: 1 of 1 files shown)


tjvr:
Hello! Thanks for working on this :-)

Unfortunately I don't think we can merge it—it's quite invasive, and passing the options down to individual States makes me sad and really isn't something we want to do!

Ideally, you would combine location information with your characters/tokens before passing them to nearley, something like this:

    parser.feed(Array.from(source).map((char, index) => ({ char, index })))

Unfortunately it then becomes harder to match them in nearley, since the "tokens" are no longer equal `===` to the literals defined in your rules. :-(

Custom tokens is actually something we've been thinking about very recently -- @hardmath123 I think this is another argument for allowing user-defined marchers! This seems like a case where adding a custom test() function to each one of your literals is clearly less convenient than supplying a custom match function.

tjvr:
Out of curiousity—what are you using position for in your postprocessors? :-)

EricMCornelius:
Returning full token location information in each AST node - which gives the ability to easily do semantic highlighting of various node types in the original input string.

As far as passing options through, the easiest (read HACKY) way to achieve this functionality was just to thread the location through and pass it as a 4th argument to the postprocessor functions.  However, the resulting signature:

```(data, start, reject, end) => {}```

Leaves a bit to be desired.

Have you guys addressed any ideas of how to pass a global context through to the postprocessors?  Seems like a very useful thing to be able to support w/o having to resort to arbitrary globals.  That was my motivation for trying to add options references, but usually I'd see that sort of thing achieved by a closure context instead of needing to explicitly assign a reference to each object type.

**Comment by tjvr** [comment]
> how to pass a global context through to the postprocessors

Smells like #63. That might be something we should revisit.

> (data, start, reject, end) => {}

Hold on; I'd forgotten postprocessors already take `start` as an argument! Given that's the case, you can work out the end index yourself as part of your AST building :-)

**Comment by EricMCornelius** [comment]
It's not possible, afaics, to calculate the end index at each node without requiring that you return that information from *every* parsed terminal - which is a *major* performance hit/grammar complexity when dealing with discarded whitespace terminals.

**Comment by tjvr** [comment]
Have you considered using a tokenizer?

**Comment by EricMCornelius** [comment]
In stricter languages I certainly prefer to, but token boundary classification would be fairly challenging for this particular DSL, unfortunately.

**Comment by tjvr** [comment]
@hardmath123 How about in postprocessors binding `this` to the Parser? (Or making the Parser the fourth argument?) Massive potential for abuse, of course; but it would stop us from adding a potentially unbounded number of arguments to State :-)

It would trivially solve @EricMCornelius's problem; you could just read`this.current`.

It might even make the people who want `context` happy...

**Comment by EricMCornelius** [comment]
As an aside, the PEG.js semantic predicates provide the full location information for rule applications, which is what I've used in the past / am most familiar with: https://github.com/pegjs/pegjs#--predicate-

It certainly does make quality error messages significantly easier to generate when you want to do (lazy) context-sensitive error generation during parsing (undefined variable names, etc.)

**Comment by tjvr** [comment]
> It certainly does make quality error messages significantly easier to generate when you want to do (lazy) context-sensitive error generation during parsing (undefined variable names, etc.)

Nearley doesn't allow giving a rejection reason, since many postprocessors may run or parsing may even continue after a rejection—see #195

**Comment by EricMCornelius** [comment]
I'd argue that valid production matches with invalid context are great times to generate quality errors. Undefined variable names being one of the canonical examples.

That being said, it's off of the main topic, which is that there's definitely value in having the end positions of production matches available, if only for applications like syntax highlighting and making semantic modifications to original input strings.

**Comment by Hardmath123** [comment]
I might be misunderstanding, but can't you get the "end" offset by subtracting 1 from the "start" offset of the _next_ item in your AST?

**Comment by EricMCornelius** [comment]
I don't want to clutter up the AST generating nodes for whitespace and other punctuation in this particular case (operators, parentheses, etc.), and whitespace could be arbitrarily long. This makes it impossible to determine which section of the input was actually matched by a given production rule in a general manner.

**Comment by EricMCornelius** [comment]
Example:
```field1 = value1
AND (field2=value2  || field3=value3)
```
Here I'd want to have an AST representing the boolean expression tree, with leaf nodes representing comparison operations.  Don't need to know location information about the parentheses or the arbitrarily structured whitespace components.

tjvr:
I'm closing this since we're not going to merge it; however you're welcome to continue the discussion here or on my proposed solution here: #203

EricMCornelius:
Yep.

Issue #188, opened 2017-03-04
nearleyc errors


Hardmath123:
If nearleyc throws a parse error when parsing an input `.ne` file, it should print out the line/col where the error happened. The current error message is confusing and messy; it's also useless when trying to locate the error in an editor.

tjvr:
Should the default error be "no possible parsings" or "invalid syntax" or "unexpected token" or something else entirely?

tjvr:
(n.b. the lexer spits out "invalid syntax" if it sees some characters it can't turn into a token, so ideally the parser error would be something different)

Hardmath123:
I'll leave lexer errors to you, but for the parser something like "Invalid syntax" should be fine. We expect people to catch this error and do something meaningful, anyway.

Hardmath123:
Also, to clarify, this issue was meant specifically for nearleyc parsing input `.ne` files. Of course, a general solution would be better, but `.ne` files are a priority because it's easier to debug syntax errors if you, y'know, _know where they are_. :P

tjvr:
> this issue was meant specifically for nearleyc parsing input .ne files

I realise; but the easiest thing is to handle this in a general way, as part of #220. :-)

tjvr:
(If you want to clean up the output, so `nearleyc` users don't see a traceback as well, that's up to you.)

Hardmath123:
> Error: invalid syntax at line 5 col 19: 
> y ->    foo %lasdf.asdf

:D

wizzard0:
Please keep the offset / token index too ;)

PR #211, opened 2017-03-23
better error in .feed method

grandsong:
When a "No possible parsings" error is thrown, now it shows the line
number and column number where the character in question is (instead of
its position in the whole chunk), as well as the content of that line.

Also changed the wording a bit.

Note: As to Line 104, the tailing whitespace was automatically deleted
by my code editor...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ test/whitespace.js
 npm-debug.log
 tmp.*
 sandbox/
-yarn.lock
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ test/whitespace.js
 npm-debug.log
 tmp.*
 sandbox/
+yarn.lock
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -101,7 +101,7 @@ Column.prototype.process = function(nextColumn) {
 
                 // special-case nullables
                 if (state.reference === this.index) {
-                    // make sure future predictors of this rule get completed.
+                    // make sure future predictors of this rule get completed. 
                     var exp = state.rule.name;
                     (this.completed[exp] = this.completed[exp] || []).push(state);
                 }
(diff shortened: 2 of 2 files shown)


tjvr:
Hi! Thanks for contributing.

There appears to be some changes missing here? Unfortunately I don't think we will want to merge this anyway; implementing this efficiently and in an input-agnostic way (Nearley supports custom tokens too!) will be quite subtle.

grandsong:
Very sorry.
I failed to include the "bin/nearley.js" which I had edited when I first commited.
And I then had a struggle with GitHub client. So now there are six commits...

Anyway, you can see my changes in "bin/nearley.js" now.


BTW, I use yarn to manage packages. I added "yarn.lock" in ".gitignore" and accidentally included this change in this RP, too.
Sorry. However, I guess sooner or later "yarn.lock" will be added since yarn is really hot nowadays.

tjvr:
Very sorry but as explained above I don't think we can merge this.

Thank you for your time :)

grandsong:
I don't quite understand you by "input-agnostic way" and "custom tokens".

How are they relative to this issue about lack of info in errors?

Were you really talking about THIS PR of mine exactly?


I am a new comer. Nearley impressed me and I believe it will be very efficient to debug my parser.

I began my learning with testing several lines of lua against the lua.ne. It works well until I added "for i == 2, n do". The "No possible parsings" error occured.

It simply shown me ONE character and its position in the chunk (as a whole bulk of string!).

That unfriendly error dampened my enthusiasm.

I don't like what I imagined: days later, when I deal with many lines of text, I get such a simple error which can help me very little to find out where the hell that character in question is!

I think the deficiency would trouble many users like me, so my improvement can be rather a necesity.

Why decline my PR and keep the errors as bad as they used to be?

tjvr:
I'm very glad you like Nearley, and thank you for contributing :) I'll try and explain my concerns more fully.

Nearley doesn't require your input to be a string; you can give `feed()` a list of token objects, for example. So looking for newlines in this way won't work for everyone.

We do agree that the error messages need improving! (See #188.) But line numbers are probably something we should handle at an earlier tokenization stage, not in the Earley parser itself.

I hope that helps explain things. Sorry for the confusion :)

grandsong:
I see.

I may come back to this topic when I learn more about other kinds of input to feed().

> "line numbers are probably something we should handle at an earlier tokenization stage"

Does that mean someday there can be a solution out-of-box for new comers like me? Or Nearley will never have line number feature?

tjvr:
> Does that mean someday there can be a solution out-of-box for new comers like me?

Yes, have a look at #188—we're planning for nearleyc errors to include line numbers.

grandsong:
Great!

Hardmath123:
Just wanted to weigh in — I'm glad you are enjoying nearley, and I thank you very much for contributing. As Tim said, we're working on implementing this functionality in a more general, robust way. Until then, keep in mind that you can convert the "offset" of the character to a line number pretty easily yourself! :-)

grandsong:
Thanks for your additional feedback.
My improvement works well for me so far and thus I will keep it for myself.
Still, looking forwards to your official solutions.

PR #231, opened 2017-05-13
Bugfix/ts preprocessor exports

hhornbacher:
TS compiler isn't complaining anymore

PR #233, opened 2017-05-15
Added Lexer type signature for TypeScript

heatherleaf:
Nice addition with lexers! TypeScript needs a signature though.


Hardmath123:
Hmm, if I'm not mistaken, I think #231 contains this fix, along with some other changes. In that case, let's wait for @hhornbacher's updated PR.

heatherleaf:
Ha! Looks like so -- I can wait :)

PR #239, opened 2017-05-25
Fixed sstrchar definition

Nopik:
Original did not properly matching code like `'foo''bar'`.

### DIFF
--- a/builtin/string.ne
+++ b/builtin/string.ne
@@ -12,7 +12,7 @@ dstrchar -> [^\\"\n] {% id %}
     }
 %}
 
-sstrchar -> [^\\\n] {% id %}
+sstrchar -> [^\\'\n] {% id %}
     | "\\" strescape
         {% function(d) { return JSON.parse("\""+d.join("")+"\""); } %}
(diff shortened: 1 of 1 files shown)


tjvr:
I don't think this is supposed to match `'foo''bar'`? @Hardmath123

Nopik:
Without the missing character, it parses `'foo''bar'` as ambiguous and produces two alternatives. If your grammar has a list of strings `fields -> field | field fields` where `field -> sqstring` then `'foo''bar'` is ambiguous with 2 alternatives: `[ 'foo\'\'bar' ]` & `[ 'foo', 'bar' ]`.

Hardmath123:
Yeah, @Nopik is right — I'm pretty sure this is a bug and that you have the right fix. Thanks for the patch!

Nopik:
Thanks. When it will arrive on npm?

Hardmath123:
Whoops, forgot to push. 2.9.3 should have the fix.

Nopik:
Great, it seems to be working fine. Thanks!

PR #234, opened 2017-05-15
Bugfix/ts preprocessor exports

hhornbacher:
My first fix was just about the typing of the datastructures and didn't fix the #223.
But I managed to fix it, there was missing the arrpush template in generate.typescript.builtinPostprocessors (generator.js)

# Test/Benchmark results before changes:
```
$ npm test

> nearley@2.9.2 test /home/XXX/Development/Private/Web/nearley
> mocha test/launch.js



  nearleyc
    ✓ should build test parser (check integrity) (130ms)
    ✓ should build for CoffeeScript (308ms)
    ✓ calculator example (44ms)
    ✓ csscolor example (60ms)
    ✓ exponential whitespace bug
    ✓ nullable whitespace bug
    ✓ percent bug
    ✓ tokens
    ✓ leo bug
    ✓ json example compiles
    ✓ json test1
    ✓ json test2
    ✓ tosh example (390ms)
    ✓ classic crontab (83ms)
    ✓ parentheses

  Parser
    ✓ shows line number in errors
    ✓ shows token index in errors
    ✓ can save state
    ✓ can rewind
    ✓ won't rewind without `keepHistory` option
    ✓ restores line numbers
    ✓ restores column number


  22 passing (1s)

$ npm run benchmark

> nearley@2.9.2 benchmark /home/XXX/Development/Private/Web/nearley
> node test/benchmark.js

  ✔ calculator arithmetic2 x 2,732 ops/sec ±1.09% (86 runs sampled)
  ✔ json sample1k.json x 34.97 ops/sec ±4.42% (47 runs sampled)
  ✔ tosh ex1 x 520 ops/sec ±1.49% (84 runs sampled)
```
# Test/Benchmark results after changes:
```
$ npm test

> nearley@2.9.2 test /home/XXX/Development/Private/Web/nearley
> mocha test/launch.js



  nearleyc
    ✓ should build test parser (check integrity) (130ms)
    ✓ should build for CoffeeScript (297ms)
    ✓ calculator example (45ms)
    ✓ csscolor example (57ms)
    ✓ exponential whitespace bug
    ✓ nullable whitespace bug
    ✓ percent bug
    ✓ tokens
    ✓ leo bug
    ✓ json example compiles
    ✓ json test1
    ✓ json test2
    ✓ tosh example (378ms)
    ✓ classic crontab (72ms)
    ✓ parentheses

  Parser
    ✓ shows line number in errors
    ✓ shows token index in errors
    ✓ can save state
    ✓ can rewind
[...]

PR #242, opened 2017-06-03
Fix TypeScript postprocessor: improve and export types

### LINKED ISSUE #223: @preprocessor typescript doesn't work with EBNF (:+, :*, etc)
When using a .ne-file like this:

```
@preprocessor typescript

test -> [a-z]:+
```

then this error occurs:

```
[...]\node_modules\nearley\lib\generate.js:18
    var lines = func.toString().split(/\n/);
                    ^

TypeError: Cannot read property 'toString' of undefined
    at dedentFunc ([...]\node_modules\nearley\lib\generate.js:18:21)
    at serializeRule ([...]\node_modules\nearley\lib\generate.js:92:53)
    at [...]\node_modules\nearley\lib\generate.js:13:14
    at Array.map (native)
    at serializeRules ([...]\node_modules\nearley\lib\generate.js:12:30)
    at Function.generate.ts.generate.typescript ([...]\node_modules\nearley\lib\generate.js:178:37)
    at generate ([...]\node_modules\nearley\lib\generate.js:107:48)
    at StreamWrapper.<anonymous> ([...]\node_modules\nearley\bin\nearleyc.js:52:22)
    at emitNone (events.js:91:20)
    at StreamWrapper.emit (events.js:185:7)
```

Tested on Node.js 6.4.0 and 7.9.0 with NPM 4.0.5 and 4.5.0 on Windows 7 x64

### LINKED ISSUE #241: Custom lexer + TypeScript = "Cannot find name" errors when compiling to JS
- `node -v`: 7.10.0
- `npm -v`: 5.0.1
- `nearleyc -v`:  2.9.2 checked out from #234 (`b530e99`)

## Problem description

Nearley compiles `%foo` into TypeScript `lexer.has("foo") ? {type: "foo"} : foo`, and that `foo` at the end is not a known name. So when I compile TS to JS, I get this error: `error TS2304: Cannot find name 'foo'`.
I'm using [Moo lexer](https://github.com/tjvr/moo) and it provides the tokens, so `lexer.has(...)` is true.

TS does get compiled to JS, but the errors lead to failed CI builds. I could try and disable all errors for this one file, but obviously it's better to have this resolved upstream.

## Minimal example

- [`example.ne`](https://github.com/deltaidea/nearley-type-definitions-example/blob/master/example.ne)
- [`example.ts`](https://github.com/deltaidea/nearley-type-definitions-example/blob/master/example.ts) - compiled from `.ne`
- [`example.js`](https://github.com/deltaidea/nearley-type-definitions-example/blob/master/example.js) - compiled from `.ts`
- [`package.json`](https://github.com/deltaidea/nearley-type-definitions-example/blob/master/package.json)

## Proposed solution

I realize that there's no easy fix for this because it's not possible to infer at compile time which custom tokens will be defined at run time and which will be provided by a lexer.

The best backwards-compatible path is to add `declare var foo: any` at the top for every custom token name. This will silence the errors, even though they are kinda valid.

I can submit a PR if this is an acceptable solution.

deltaidea:
Mega-PR with all the fixes for TypeScript.

- Improve and export interfaces from the generated TS. Closes #233, #234.
- Add `arrpush` built-in postprocessor as in other generators. Closes #223.
- Declare custom tokens in the generated TS to avoid errors when compiling to JS. Closes #241.

The first commit (055d79bc213102173118f67b1078c0f032066c21) is a squashed and consistently formatted version of #234. I double-checked, it's equivalent to that PR, nothing was lost. That PR looks larger because of the formatting changes.

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -15,6 +15,7 @@ function Compile(structure, opts) {
     var result = {
         rules: [],
         body: [], // @directives list
+        customTokens: [], // %tokens
         config: {}, // @config value
         macros: {},
         start: ''
@@ -52,6 +53,7 @@ function Compile(structure, opts) {
                 require('./lint.js')(c, {out: process.stderr});
                 result.rules = result.rules.concat(c.rules);
                 result.body  = result.body.concat(c.body);
+                result.customTokens = result.customTokens.concat(c.customTokens);
                 Object.keys(c.config).forEach(function(k) {
                     result.config[k] = c.config[k];
                 });
@@ -126,6 +128,9 @@ function Compile(structure, opts) {
         if (token.token) {
             if (result.config.lexer) {
                 var name = token.token;
+                if (result.customTokens.indexOf(name) === -1) {
+                    result.customTokens.push(name);
+                }
                 var expr = result.config.lexer + ".has(" + JSON.stringify(name) + ") ? {type: " + JSON.stringify(name) + "} : " + name;
                 return {token: "(" + expr + ")"};
             }
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -173,7 +173,7 @@ generate.ts = generate.typescript = function (parser, exportName) {
     output += parser.customTokens.map(function (token) { return "declare var " + token + ":any;\n" }).join("")
     output += parser.body.join('\n');
     output += "export interface Token {value:any; [key: string]:any};\n";
-    output += "export interface Lexer {reset:(chunk:string, info:any) => void; next:() => Token; save:() => any; formatError:(token:Token) => string; has:(tokenType:string) => boolean};\n";
+    output += "export interface Lexer {reset:(chunk:string, info:any) => void; next:() => Token | undefined; save:() => any; formatError:(token:Token) => string; has:(tokenType:string) => boolean};\n";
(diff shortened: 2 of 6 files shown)


deltaidea:
This is blocked until DefinitelyTyped/DefinitelyTyped#16937 is merged. Until then, `npm install` fails.

tjvr:
This looks great at first glance, thank you! I like that you added tests :-)

@hardmath123 do you have time to review this at the moment?

Hardmath123:
I have time, but I don't think I'm qualified to. I don't really speak Typescript.

deltaidea:
Everything's ready here. Just checked, and that PR has been merged and `npm test` now succeeds.

Hardmath123:
Okay, I merged it. Version 2.10.0 is live on npm.

deltaidea:
Thank you! I'm sorry this is left without a proper review. Is there a way to provide a custom generator like this one from the outside? That would allow to separate TS into its own repo, and you wouldn't have to support it officially.

Hardmath123:
Not quite, no, but that's not a bad idea. @tjvr ?

tjvr:
@Hardmath123 If you're going to do *that*, then perhaps we want nearleyc to be its own package? Regardless, a broader refactor of generate/compile might be useful at some point.

PR #245, opened 2017-06-17
Strict mode compatibility fix

simonhildebrandt:
I pulled [the awesome!] Nearley into a personal project via Browserify and found that when Browserify added 'use strict', Nearley crashed in a `where` clause. This change fixes it (and I tweaked the docs as well. :) )


tjvr:
Great—thank you! :D

PR #254, opened 2017-07-30
[WIP] Improve the docs

deltaidea:
# [Preview of the progress so far](https://github.com/deltaidea/nearley/blob/improve-docs/README.md)

This is a joint effort to make the docs more straight-forward and complete.

I hijacked `tjvr:ihatewriting` branch and started this PR based on it. @tjvr can still push to this PR since he's a collaborator, and I didn't want to deal with plain-text conflicts.

Relevant:

- [Art of Readme](https://github.com/noffle/art-of-readme)
- [Standard Readme](https://github.com/RichardLitt/standard-readme)

Things to do (in a vaguely sensible order):

- [x] Reorder readme, reword introduction. It's awkward for beginners.
- [x] Update all code examples to ES6.
- [x] Document `Grammar.fromCompiled()` properly. (#247, #248, #257)
- [x] Make sure `{% id %}` is documented high enough that you can't miss it. (#244)
- [x] Explain how to use in the browser. (#142, #246, #256)
- [x] Add recipes in hope of saving users' time and consistency in the community:
    - [x] Clean example of using [moo](https://github.com/tjvr/moo).
    - [x] Minimal example of an indentation-aware language. (#251)
    - [x] Generating CST by returning objects from postprocessors. (#258)
- [x] Add a note that CoffeeScript and TypeScript are supported. (#252)
- [ ] Performance tips: avoiding ambiguity, using moo. (#81, #238, #243, #250)
- [ ] Error handling:
    - [ ] List of possible exceptions. I imagine it's short, but I'm not sure about it myself.
    - [ ] Empty `parser.results`. (#253)
- [ ] Debugging tips: inspecting that the right tokens are parsed, using `nearley-test`. (#253)
- [ ] IDE support: syntax highlighting is supported at least in VS Code, probably in other editors too.
- [ ] Go through the source code and update references to the readme.
- [x] Don't forget to run `doctoc --notitle README.md` before merging.

PR #259, opened 2017-08-04
Add ES2015 (ES6) generator

darrylyeo:
I didn't like how the generated ES5 code looked in my ES2015 project, so I thought I'd make myself a generator.

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -156,8 +156,8 @@ generate.es2015 = generate.es6 = function (parser, exportName) {
 };
 generate.es2015.builtinPostprocessors = {
     "joiner": "d => d.join('')",
-    "arrconcat": "d => [d[0]].concat(d[1])",
-    "arrpush": "d => d[0].concat([d[1]])",
+    "arrconcat": "([el, arr]) => [el].concat(arr)",
+    "arrpush": "([arr, el]) => arr.concat([el])",
     "nuller": "_ => null",
     "id": "id"
 }
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -135,6 +135,33 @@ generate.javascript.builtinPostprocessors = {
     "id": "id"
 }
 
+
+generate.es2015 = generate.es6 = function (parser, exportName) {
+    return "// Generated automatically by nearley\n" +
+        "// http://github.com/Hardmath123/nearley\n" +
+        "{\n" +
+        "const id = x => x[0];\n" +
+        parser.body.join('\n') +
+        "const grammar = {\n" +
+        "    Lexer: " + parser.config.lexer + ",\n" +
+        "    ParserRules: " + serializeRules(parser.rules, generate.es2015.builtinPostprocessors) + ",\n" +
+        "    ParserStart: " + JSON.stringify(parser.start) + "\n" +
+        "}\n" +
+        "if (typeof module !== 'undefined' && typeof module.exports !== 'undefined') {\n" +
+        "    module.exports = grammar;\n" +
+        "} else {\n" +
+        "    window." + exportName + " = grammar;\n" +
+        "}\n" +
+        "}\n";
+};
+generate.es2015.builtinPostprocessors = {
+    "joiner": "d => d.join('')",
+    "arrconcat": "d => [d[0]].concat(d[1])",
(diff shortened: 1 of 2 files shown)


lib/generate.js
```diff
@@ -135,6 +135,33 @@ generate.javascript.builtinPostprocessors = {
...
+    return "// Generated automatically by nearley\n" +
+        "// http://github.com/Hardmath123/nearley\n" +
+        "{\n" +
+        "const id = x => x[0];\n" +
+        parser.body.join('\n') +
+        "const grammar = {\n" +
+        "    Lexer: " + parser.config.lexer + ",\n" +
+        "    ParserRules: " + serializeRules(parser.rules, generate.es2015.builtinPostprocessors) + ",\n" +
+        "    ParserStart: " + JSON.stringify(parser.start) + "\n" +
+        "}\n" +
```
deltaidea: Missing `;`

lib/generate.js
```diff
@@ -135,6 +135,33 @@ generate.javascript.builtinPostprocessors = {
...
+        "    module.exports = grammar;\n" +
+        "} else {\n" +
+        "    window." + exportName + " = grammar;\n" +
+        "}\n" +
+        "}\n";
+};
+generate.es2015.builtinPostprocessors = {
+    "joiner": "d => d.join('')",
+    "arrconcat": "d => [d[0]].concat(d[1])",
+    "arrpush": "d => d[0].concat([d[1]])",
```
deltaidea: Could've done `([el, arr]) => [el].concat(arr)` and `([arr, el]) => arr.concat([el])`

lib/generate.js
darrylyeo: Heh. Side effect of copying from the regular JS generator (see line 120).

If it's all right with you, I'd love to make another PR to make the coding style consistent.

lib/generate.js
darrylyeo: Ooh, good idea.

lib/generate.js
deltaidea: I'm just a fellow contributor, but sure! There're no other PRs to the codebase at the moment, so no potential conflicts.

It may be a good idea to make an issue first and discuss code styleguide properly, because there's none and the style is all over the spectrum. If you'd champion that issue, I'd be a happy man. I currently only have time to slowly inch forward with the docs over at #254.

tjvr:
Thanks for getting involved in Nearley! :-)

I'm afraid I don't think we're going to merge this PR. The generated code isn't meant to be read, and it has no benefit apart from aesthetics to keep around an ES6 version

Regarding updating the code style of the project: we'd like to stick with ES5/UMD for compatibility, performance, and to avoid transpilers such as Babel.

Personally I'd prefer a style that dropped semicolons (like how [moo](tjvr/moo) is written), but it's not particularly worth arguing about: there are more important issues!

We at least want to stick with ES5 for the node core (ie. `nearley.js`). The tests already target Node 6 / ES6, and so it's plausible that the surrounding tooling—which only targets node—could use ES6. But I'm not sure @hardmath123 is ready for that yet :-)

independentgeorge:
@tjvr So there is currently no way to use the generated grammar and `nearley.js` from inside an ES6 project, and there are no plans to add support for it?

tjvr:
> there is currently no way to use the generated grammar and nearley.js from inside an ES6 project

That's not the case at all! ES6 is, of course, backwards-compatible with ES5; so you can use the generated grammar and nearley.js as normal. :-)

independentgeorge:
By ES6 I meant the use of ES modules. I'm using [rollup](https://github.com/rollup/rollup) in my project and the UMD wrappers are getting in the way of `import`s.

I'm asking about support for importing `nearley.js` directly as an ES6 module, and specifying the output format of the generated grammar to ES6 modules so they could be imported as well.

I know this is not related to the PR, and I apologise if I'm asking in the wrong place, but you did mention some of what I'm talking about.

tjvr:
I see. That probably should be a separate issue, yes! But since we’re here, I’ll reply.

I don’t think we’re going to support native `import`, at least not for a good while. Surely Rollup must provide backwards-compatibility with UMD modules... if not, then I don’t see it as nearley’s fault :-)

PR #260, opened 2017-08-09
Adds link to nearley-gulp in README

JosephJNK:
I wrote a small gulp plugin wrapper around the existing `nearly-loader` plugin, and thought it might be useful to others as well. I tried to match the format of the other tooling links in the README.


Hardmath123:
Nice! Thank you.

PR #266, opened 2017-08-09
Move lib & refactor tests

tjvr:
This:

* Splits out `nearley.js` (the core parsing algorithm) from `nearleyc` (the grammar compiler/generator)
* Refactors the tests: uses `jest` instead of `mocha` & `chai` (to get snapshots, async, expect)
* Adds a new script (`npm run watch`) for re-running the tests on save


Hardmath123:
…why?

tjvr:
This aids documentation, as well as properly separating the core parsing algorithm from the grammar specification language.

tjvr:
I'm not sure what you're upset about here. Can you be more specific? :)

Hardmath123:
I'm not upset! I just like to know why we're making changes, because a solid 40% of nearley problems are people making changes for the sake of making changes and breaking something a few months later.

deltaidea:
This is a step in the right direction, I'd say, but this will break a shitton of integrations. People require internal stuff, e.g. `nearleyc` in a browser.

This reminds me: it's a good idea to keep a minimal changelog with migration notes, or write that down in the releases and add a note that you can read changelogs there.

Hardmath123:
> shitton

Language, please. There's a small community of kids who use nearley to make fun languages. Let's keep that in mind.

tjvr:
> People require internal stuff, e.g. `nearleyc` in a browser.

But the idea of adding `nearleyc/index.js` is _so that_ it's easier to use nearleyc in a browser! The idea is that you'll be able to just do `require('nearley/nearleyc')` or whatever.

deltaidea:
Yeah, sure, I'm just saying, this will break existing code and it's a good idea to notify somehow that this is the case.

Hardmath123:
does this break the quick compiler in `/www`?

tjvr:
Yes! Fixed in 331147c.

tjvr:
So. I still think this is a reasonable change. If people are depending on the internal structure of the package, then that's something that we're going to break sooner or later. We might as well make it sooner, and as part of this PR come up with a _supported_ way for using nearleyc in a browser.

Thoughts on the API for such a thing would be appreciated. Perhaps `require('nearley/nearleyc')` returns a function, which takes a `ne` string and returns the generated source? Then you can use a third-party lib like `require-from-string` to turn it into a grammar you can pass to `nearley.Grammar.fromCompiled`. But that already seems too convoluted... :-)

deltaidea:
Don't forget to update `docs/using-in-frontend.md`!

tjvr:
@deltaidea Absolutely! Once we've decided what the API should look like ;-)

tjvr:
Closing, we don't want to break backwards compat (at least not right now!).

Splitting this into #267 and #268. L-)

PR #269, opened 2017-08-17
Refactor benchmarks using benchr

tjvr:
This halves the amount of code in `benchmarks.js`, makes it easier to compare benchmarks across different libraries if we want to do that, and makes it easier to add new benchmarks. :-)


Hardmath123:
Thanks Tim!

PR #271, opened 2017-08-18
Optimise docs site

tjvr:
This avoids FOIT in favour of FOUT: that is, show a fallback font while the webfont (Roboto) is loading. This means the content is readable in half the time, on a slow connection. (This is how `@font-face` used to work, before browsers changed their behaviour.)

This also inlines the CSS, to avoid an additional blocking request before the content can be rendered.

Before (>7s to content):
screen shot 2017-08-18 at 13 42 49

---
After: (<3s to content):
screen shot 2017-08-18 at 13 43 24


Hardmath123:
Thanks Tim!

PR #272, opened 2017-08-18
Split up docs/index some more

tjvr:
Yo @Hardmath123. This splits out all the main index sections into their own chapters -- and then pulls in related bits, to fill out each chapter a bit more.

Have a look at the commits for more details. :-)


Hardmath123:
Oh no! Merge conflicts!

Hardmath123:
Perhaps the files should be `compiler.html` and `parser.html` instead of "language" and "usage", which are kind of confusing?

tjvr:
I don't like the name "compiler" (I think it gets confusing calling `nearleyc` a compiler). Perhaps `ne-syntax` and `parser`?

Hardmath123:
> I think it gets confusing calling nearleyc a compiler

Yeah, even though that's really what it is — a syntactic-sugar-expander. It's confusing because we use a grammar to make a parser to parse a grammar for grammars that make other parsers.

Maybe `grammar` and `parser`?

tjvr:
Merged & renamed!

PR #281, opened 2017-08-25
docs: change pages root to /docs

tjvr:
@Hardmath123 I think you had some changes to push to this.


Hardmath123:
I'm just going to start over on this, since we decided we wanted to keep `/docs/` as part of the URLs.

tjvr:
ok fam

PR #283, opened 2017-08-25
Split tests up

tjvr:
To make 'em easier to manage. (I'm about to add some more tests for Leo stuff, you see.)


Hardmath123:
> To make 'em easier to manage.

Okay, will trust you on that.

Issue #208, opened 2017-03-21
Update Dependencies


tjvr:
- [x] Upgrade mocha 5319f7a
- [ ] nomnom is unmaintained; switch to [commander](https://github.com/tj/commander.js)

tjvr:
It turns out `commander` isn't just a drop-in replacement for `nomnom`; I hadn't investigated past the initial warning message.

PR #284, opened 2017-08-26
Ditch deprecated nomnom in favor of commander

stefanmaric:
I tried to edit as little as possible. Some minor things to notice:

* commander's generated help has no color, unlike nomnom's
* commander has no positional elements, that means:
  - [file] can actually be passed last
  - but CLI usage remains backwards compatible
* I manually tested the all the CLI commands and options but it would
  be better if someone double checks.

Part of #208

### DIFF
--- a/bin/nearley-railroad.js
+++ b/bin/nearley-railroad.js
@@ -10,26 +10,19 @@ try {
 
 var fs = require('fs');
 var path = require('path');
-var nomnom = require('nomnom');
+var opts = require('commander');
 
-var opts = nomnom
-    .script('nearley-railroad')
-    .option('file', {
-        position: 0,
-        help: "A grammar .ne file (default stdin)"
-    })
-    .option('out', {
-        abbr: 'o',
-        help: "File to output to (default stdout)."
-    })
-    .option('version', {
-        abbr: 'v',
-        flag: true,
-        help: "Print version and exit",
-        callback: function() {
-            return require('../package.json').version;
-        }
-    }).parse();
+opts
+    .description('Generate pretty railroad diagrams from a parser')
+    .usage('[file] [options]')
+    .option(
+        '-o, --out <file>',
+        'File to output to (defaults to stdout)'
+    )
+    .version(require('../package.json').version)
+    .parse(process.argv);
+
+opts.file = opts.args[0]
 
 var input = opts.file ? fs.createReadStream(opts.file) : process.stdin;
 var output = opts.out ? fs.createWriteStream(opts.out) : process.stdout;
--- a/bin/nearley-test.js
+++ b/bin/nearley-test.js
@@ -6,43 +6,33 @@
 
 var fs = require('fs');
 var nearley = require('../lib/nearley.js');
-var nomnom = require('nomnom');
+var opts = require('commander');
 var StreamWrapper = require('../lib/stream.js');
 
-var opts = nomnom
-    .script('nearley-test')
-    .option('file', {
-        position: 0,
-        help: "A grammar .js file",
-        required: true,
-    })
-    .option('input', {
-        abbr: 'i',
-        help: "An input string to parse (if not provided then read from stdin)",
-        type: 'string',
-    })
-    .option('start', {
-        abbr: 's',
-        help: "An optional start symbol (if not provided then use the parser start symbol)",
-    })
-    .option('out', {
-        abbr: 'o',
-        help: "File to output to (defaults to stdout)",
-    })
-    .option('quiet', {
-        abbr: 'q',
-        flag: true,
-        help: "Output parse results only (hide Earley table)",
-    })
-    .option('version', {
-        abbr: 'v',
-        flag: true,
-        help: "Print version and exit",
-        callback: function() {
-            return require('../package.json').version;
-        }
-    })
-    .parse();
+opts
+    .description('Test a grammar against some input and see the results')
+    .usage('[file] [options]')
+    .option(
+        '-i, --input <file>',
+        'An input string to parse (if not provided then read from stdin)'
+    )
+    .option(
+        '-s, --start <symbol>',
+        'An optional start symbol (if not provided then use the parser start symbol)'
+    )
+    .option(
+        '-o, --out <file>',
+        'File to output to (defaults to stdout)'
+    )
+    .option(
+        '-q, --quiet',
+        'Output parse results only (hide Earley table)',
+        false
+    )
+    .version(require('../package.json').version)
+    .parse(process.argv);
+
+opts.file = opts.args[0]
 
 var output = opts.out ? fs.createWriteStream(opts.out) : process.stdout;
 
--- a/bin/nearley-unparse.js
+++ b/bin/nearley-unparse.js
@@ -2,43 +2,40 @@
 
 var fs = require('fs');
 var nearley = require('../lib/nearley.js');
-var nomnom = require('nomnom');
+var opts = require('commander');
 var randexp = require('randexp');
 
-var opts = nomnom
-    .script('nearley-unparse')
-    .option('file', {
-        position: 0,
-        help: "A grammar .js file",
-        required: true,
-    })
-    .option('start', {
-        abbr: 's',
-        help: "An optional start symbol (if not provided then use the parser start symbol)",
-    })
-    .option('count', {
-        abbr: 'n',
-        help: 'The number of samples to generate (separated by \\n).',
-        default: 1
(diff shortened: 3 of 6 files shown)


Hardmath123:
I'm not sure what we gain from this PR.

Maybe in the next major nearley version we can rewrite the command-line interfaces, but for now this seems like rewriting things for the sake of replacing nomnom. nomnom has worked really well for everyone the past 4 years, so I don't see any urgency to replacing it — the worst-case scenario (breaking someone's build script) hurts more than the best-case scenario (nothing changes) helps.

In conclusion, I appreciate that you tried to change as little as possible (though line 35 on bin/nearley-test.js seems unnecessary?), but I don't think the change is needed.

Also, unless I'm misunderstanding you, the nomnom version totally lets you put [file] last!
```js
$ nearleyc --nojs <(echo 'a -> b {% buzz, buzz %}')
WARN	Undefined symbol `b` used.
// Generated automatically by nearley
// http://github.com/Hardmath123/nearley
(function () {
function id(x) {return x[0]; }
var grammar = {
    Lexer: undefined,
    ParserRules: [
    {"name": "a", "symbols": ["b"]}
]
  , ParserStart: "a"
}
if (typeof module !== 'undefined'&& typeof module.exports !== 'undefined') {
   module.exports = grammar;
} else {
   window.grammar = grammar;
}
})();
```

tjvr:
Thanks for working on this! I’ve been meaning to do this for a while. :-)

Since we’re worried about breaking changes, how about we merge this into a `v3` branch, @Hardmath123? I was about to create one (just with dropping postprocessors, so far).

stefanmaric:
> I'm not sure what we gain from this PR.

I actually was simply looking something else on the issues and saw #208 and thought "that's easy!". lol

But I guess the benefits are reducing technical debt and reducing dependency tree size (64kb vs 260kb of nomnom).  

> (though line 35 on bin/nearley-test.js seems unnecessary?

I believe it would break, `opts.file` is later used and `commander` doesn't have a way to declare positional arguments, they are just stacked on `commander.args`.

> unless I'm misunderstanding you, the nomnom version totally lets you put [file] last!

Your example is using `stdin`, but I double checked and you're right: `nomnom` allows for the [file] path to be last. I thought the positional arguments included the options, but they don't.

> how about we merge this into a v3 branch

Do you guys have a roadmap/milestone/plan for that? I would like to help. <3

stefanmaric:
> Do you guys have a roadmap/milestone/plan for that?

Nevermind, I found it: https://github.com/Hardmath123/nearley/projects/1

Hardmath123:
> I believe it would break, opts.file is later used and commander doesn't have a way to declare positional arguments, they are just stacked on commander.args.

Ew.

> Since we’re worried about breaking changes, how about we merge this into a v3 branch

No. We're not merging this. I needed to draw a line _somewhere_, and apparently this is where it gets drawn.

Look, I get it. It's fun to architect big software projects, to choose which libraries to use, to organize directories and build systems and test frameworks and whatnot. Making such choices feels good, like scratching an itch, like ordering the perfect burrito at Chipotle. Why do you think there are so many JS frameworks all over the place?

But you need to understand that this is purely lateral movement (at best). You rip something out, replace it with something else, eliminate some quirk that maybe bothers you a little bit, and in the process introduce a new set of subtle bugs for me to deal with two years down the line. All the while thinking that you've improved things somehow. As I said, it's like Chipotle: you get a high from cheap "that's easy!. lol" [sic] calories while under the illusion that you're eating vegetables.

If I cared less about nearley, this would be no big deal. But I _do_ care. This is not a playground for anyone with an opinion about nodejs to push their agenda. This is a piece of software used by many humans, which has my name on it. I want it to be good. I want it to be easy to use. I want people to take pleasure in using it. And lateral movement does none of those things; it just sinks up my time and yours in a frantic race to meet some nebulous and unimportant ideals that _none_ of our happy users care about.

So from now on, can we please try to move _forward?_ I'll start merging PRs as soon as they start making this project _better_ instead of _different._ Until then, if you need to scratch that itch, do it elsewhere.

---

Sorry to get you caught in the cross-fire, @stefanmaric. I'm all for people contributing to open-source. But I hope you understand that as a maintainer, I shouldn't be obligated to merge every pull request? I do hope this doesn't discourage you from working on nearley if you wish to — I just encourage you to make meaningful changes (I'm happy to suggest tasks that I would actually appreciate done). :-)

> Nevermind, I found it: https://github.com/Hardmath123/nearley/projects/1

I was of course not aware of this.

tjvr:
> I was of course not aware of this.

I started playing around with that when I was talking to you about v3 the other day. I was trying to think about which things would indeed make things better. Sorry for the confusion.

stefanmaric:
> I'm happy to suggest tasks that I would actually appreciate done

No offense, that's what issues and their labels, among other tools, are for; that's why I opened the PR – I saw "Help wanted" on #208.

My 2¢: if you're concerned about the direction of the project, you should definitely remove the collaborators – that will bring control back to you.

I'm using `nearley` so I'll be probably scanning the issues time to time and would love to have the opportunity to offer help. Thanks for sharing it.

Love <3

tjvr:
> > I'm happy to suggest tasks that I would actually appreciate done
> No offense, that's what issues and their labels, among other tools, are for; that's why I opened the PR – I saw "Help wanted" on #208.

Yes, this is my fault—I saw the warning from `nomnom` and thought this would be a good idea. I was wrong!

graingert:
@tjvr is this worth revisiting now that nomnom has been replaced with a security package?

tjvr:
Hm, does this actually cause a problem for anyone? :)

graingert:
@tjvr yes, these console warnings fill me with a burning rage :fire:

nachocab:
@tjvr the only problem I've noticed is that a grammar that expects an input that starts with `-` cannot be tested. For example:

```
S -> "-"
```

`nearly-test` gives this error because `nomnom` treats dashes as options (unless they contain a space harthur/nomnom#37):

```
nearley-test ./grammar.js --input "-"

'-input' expects a value
```

graingert:
@tjvr how about spliting the package into @nearley/cli and @nearley/runtime

hulkish:
@tjvr can we get a next tagged publish of this?

graingert:
@hulkish it's not even been merged yet

hulkish:
@graingert, I can see that. I'm suggesting maybe it should be - it doesn't have to be a minor/patch release due to concerns of breaking changes. But, it can be a major bump release?

graingert:
One other option would be to ask to take over the nomnom name and
maintenance

hulkish:
Won't that just prolong the inevitable? I'm not clear on why nomnom is deprecated. Bu, it seems to me that switching to an already maintained package (commander) would be the practical choice.

Commander also [has a smaller footprint](https://github.com/kach/nearley/pull/284#issuecomment-325137246). Sounds logical to make the switch, to me.

honzajavorek:
The latest version of the `nomnom` package is deprecated, thus won't ever get any updates and bug fixes. Except for that, it has following licensing issues in its dependencies:

- ansi-styles@1.0.0 (MIT) - missing full license text
- has-color@0.1.7 (MIT) - missing full license text
- strip-ansi@0.1.1 (MIT) - missing full license text

The MIT license explicitly requires to attach the full license text to the code or any distribution. Without this the license isn't fulfilled and thus invalid. Unlicensed code, as distributed in the npm packages mentioned above, is technically a proprietary code nobody can legally use. `nearly` is dynamically linking these 4th party dependencies due to its dependency on `nomnom`, which renders `neraly` unusable in any environment where people actually do care about licensing.

One could upgrade `ansi-styles` to `ansi-styles@3.2.1`, which has licensing in order, but guess what - `nomnom` won't get any updates or fixes. The same goes with `has-color@0.2.0`, or with `strip-ansi@4.0.0`.

Is this a good enough reason for you to migrate to a different CLI solution? Given there are alternatives, such as `commander` and `yargs`, which have all their dependencies legally OK, and there is even this PR, which implements the transition...

patricksmms:
So there are really good reasons to replace `nomnom` as stated above. Can this please be revisited? @kach

opichals:
Perhaps just changing the package.json to use one of the non-deprecated fork (e.g. this `@gerhobbelt/nomnom`) would work here?

honzajavorek:
That's what they consider in Facebook as well: https://github.com/facebook/jscodeshift/pull/240

kach:
Hopefully fixed in 2.16.0.

PR #287, opened 2017-08-27
Potential TypeScript naming collisions

alexandertrefz:
### The Problem
This PR aims to remove common naming collisions from the TypeScript code generation for the following reason:

With the current implementation, code like this fails in a weird way:

```
@preprocessor typescript
@{%
import { Lexer } from '../lexer'
const lexer = new Lexer()
%}

@lexer lexer
```

Typescript will  tell us 2 warnings:
1) `TS2440: Import declaration conflicts with local declaration of 'Lexer'.`
2) `TS2351: Cannot use 'new' with an expression whose type lacks a call or construct signature.`

This is because this code translates to Typescript that lives in the same scope as the boilerplate generation nearley does at the end of the file, which currently contains `export Lexer: Lexer | undefined = ...` - which is hard to understand unless you inspect the entire `grammar.ts` output file in detail.

So Typescript attempts to compile anyway and ignores our import and assumes that `Lexer` is meant to be the last definition, which lives on `exports` and is the assignment of the instance we attempted to create in the first place. To remedy this, I have done this PR.

### The Solution

This PR changes the Typescript Code Generation in 2 significant ways:

- it renames 2 interfaces
- it changes the exports to a single `grammar` object, instead of 3 individual exports
  - this has 2 effects:
  - a) align with JS & CS who already export such an object
  - b) make it nicer to use the API

Example for b) new: `import { grammar } from './grammar'` vs old: `import * as grammar from './grammar'`

## Important Notes
- This PR breaks backwards compatibility for typescript users, but the needed change is minimal as shown above
- The PR adds another optional argument to `serializeRules` to make the code generation output look nicer

### DIFF
--- a/docs/md/parser.md
+++ b/docs/md/parser.md
@@ -149,6 +149,15 @@ file. This can be useful to write your postprocessors in a different language,
 and to get type annotations if you wish to use nearley in a statically typed
 dialect of JavaScript.
 
+#### Typescript Preprocessor Changes
+
+If you have recently upgraded to a newer version of nearly, you might have run
+into an issue with `error TS2345` talking about a variable
+`missing the following properties from type 'CompiledRules': ParserStart, ParserRules`
+This is a result of a change in the output format of the grammar, but is very easily fixed.
+Just change the way you import the grammar from `import * as grammar from "./path/to/grammar"`
+to `import grammar from "./path/to/grammar"` and everything should compile nicely again.
+
 ### What's next?
 
 Now that you know how to use parsers, [learn how to add a tokenizer to speed
--- a/docs/parser.html
+++ b/docs/parser.html
@@ -376,6 +376,13 @@ instead choose to compile to CoffeeScript or TypeScript by adding
 file. This can be useful to write your postprocessors in a different language,
 and to get type annotations if you wish to use nearley in a statically typed
 dialect of JavaScript.</p>
+<h4 id="typescript-preprocessor-changes">Typescript Preprocessor Changes</h4>
+<p>If you have recently upgraded to a newer version of nearly, you might have run
+into an issue with <code>error TS2345</code> talking about a variable
+<code>missing the following properties from type &#39;CompiledRules&#39;: ParserStart, ParserRules</code>
+This is a result of a change in the output format of the grammar, but is very easily fixed.
+Just change the way you import the grammar from <code>import * as grammar from &quot;./path/to/grammar&quot;</code>
+to <code>import grammar from &quot;./path/to/grammar&quot;</code> and everything should compile nicely again.</p>
 <h3 id="what-s-next-">What’s next?</h3>
 <p>Now that you know how to use parsers, <a href="tokenizers">learn how to add a tokenizer to speed
 things up!</a></p>
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -6,10 +6,14 @@
     }
 }(this, function(nearley) {
 
-    function serializeRules(rules, builtinPostprocessors) {
-        return "[\n    " + rules.map(function(rule) {
-          return serializeRule(rule, builtinPostprocessors);
-        }).join(",\n    ") + "\n]";
+    function serializeRules(rules, builtinPostprocessors, extraIndent) {
+        if (extraIndent == null) {
+            extraIndent = ''
+        }
+    
+        return '[\n    ' + rules.map(function(rule) {
+            return serializeRule(rule, builtinPostprocessors);
+        }).join(',\n    ') + '\n' + extraIndent + ']';
     }
 
     function dedentFunc(func) {
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -10,7 +10,7 @@
         if (extraIndent == null) {
             extraIndent = ''
         }
-    
+
         return '[\n    ' + rules.map(function(rule) {
             return serializeRule(rule, builtinPostprocessors);
         }).join(',\n    ') + '\n' + extraIndent + ']';
@@ -131,6 +131,7 @@
         output += "})();\n";
         return output;
     };
+
     generate.javascript.builtinPostprocessors = {
         "joiner": "function joiner(d) {return d.join('');}",
         "arrconcat": "function arrconcat(d) {return [d[0]].concat(d[1]);}",
@@ -174,6 +175,7 @@
         output += "    window." + exportName + " = grammar;\n";
         return output;
     };
+
     generate.coffeescript.builtinPostprocessors = {
         "joiner": "(d) -> d.join('')",
         "arrconcat": "(d) -> [d[0]].concat(d[1])",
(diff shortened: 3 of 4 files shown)


alexandertrefz:
The `Improve Typescript Code generation` might be a mediocre title for that particular commit - if you want me to change it to something better, let me know and I'll rebase it quickly.

Hardmath123:
Wouldn't this be more easily fixed by naming your lexer class anything _except_ `Lexer`?

I appreciate the PR, but I'm very uncomfortable about merging something that breaks backwards-compatibility for TypeScript, even if the change required for the user is minimal. There is a lot of old nearley-ts code out there — I don't want to break it in unobvious ways for anyone.

Perhaps we can come up with an alternate solution?

alexandertrefz:
I figured this would be a concern.

Technically yes, you can work around this by giving your `Lexer` a different name, but you still need to figure out why this issue even occurs, which can be quite time consuming - it took me quite a while as I have a sizeable parser definition and the 2 colliding lines are at the very start and very end of the file respectively.

I also would like to point out again that the API is not aligned with JS & CS currently as well as the fact that the needed import to get the Lexer working is also unobvious (and not very well documented, probably because of said unaligned APIs).

The only way is:

```typescript
import * as grammar from './grammar'
import { Parser, Grammar } from 'nearley'
let parser = new Parser(Grammar.fromCompiled(grammar))
```
Attempting to pass the individual variables that `./grammar` exports yields in frustration because there is no way to pass in `Lexer`.

This PR changes both issues (even though there could be done more to fix the second part, but that is a bigger effort in the documentation & codebase alignment)

So while this change sure would be annoying to old users, I would argue it is even more annoying to new users. I am also going to point out that this is what SemVer was made for, even though people use it incorrectly often enough.

`2.12.0` is not an update anybody should be upgraded to by accident, and people that update this manually should have a look in the patch notes where this should be notes.

Otherwise this could also be moved to `v3` which is going to break all sorts of compatibilities anyway, right?

Hardmath123:
I agree that the output should follow the same format/API as JS and CS, and that its interface should be documented. I just want to fix this in a way that is as painless as possible for end-users. A bad situation I imagine is:

* User forks some old nearley-typescript-based project.
* User modifies grammar.ne and compiles it with the `nearleyc` they have lying around in their $PATH, which happens to be 2.12.4 or something.
* Suddenly the project breaks and it's not at all clear why. Nothing that either user did was objectionable.

Little things like this are what make most of today's software extremely painful to use, and I am going to do my very best to avoid them in nearley.

Let's try the following approach:
* First of all, the docs for preprocessors, both CS and TS, should mention how to import the grammar "correctly". This was a silly omission on our part.
* `nearleyc` 2.x.y where x > 11 should emit a warning (via the linter?) if you compile with `@preprocessor typescript`. The warning should note that the output interface has changed since nearleyc 2.12 and explain the new way to import TypeScript-based nearley grammars. This way, TypeScript users don't get caught unaware by this change.
* The docs should mention this change.

By the way, doesn't your change now cause the same issue with `Lexer` if someone defines or imports a name called `grammar` in their nearley grammar? I see the whole naming collision thing as purely a documentation issue — the real value of this PR is getting the interface to be the same as JS/CS, since I see that biting us in the future.

Hardmath123:
Looks like it _used_ to be correct, but was "refactored" in 46bec2dc392302dc8acc4b6041e6756cad8314e5 by @deltaidea (#242). Since that was only merged in June, I'm a lot less worried about backwards-incompatibility: I don't think we'll be breaking _too_ much.

In my defense, I totally said "I don't think I'm qualified to [review this PR]. I don't really speak Typescript." We'll be more careful in the future. :-)

alexandertrefz:
I would argue that people should use a local version, declared with the dependencies of the project to compile the project that they forked(which would avoid that issue) but I understand your position.

Your solution approach sounds good.

If people create a variable grammer they could run into the same issue, though likely a slightly different one. That could be solved rather well by doing an `export default { Lexer, ParserRules, ParserStart }` rather than a named export - I didnt opt for that for this PR as that would break not only old TS grammars, but all old grammars, considering that would want to keep the API the same for all 3 languages and thus change it for all 3. Actually thinking about this more I am not sure if CoffeeScript supports that at all - so I am not sure if that would be the way to go.

I think there might be a solution where we put all the user generated code in an IIFE that returns all the things we need, which could solve the naming collision generalisable for all 3 languages, but if we use that or not, we still should change the TS API to align with the other APIs - your solution for that sounds well.

I think `nearleyc` should support something like `--silent` to hide the warning, if you want to show it and you should probably remove it after a couple of versions. But from that second comment it sounds like we might not need to show it at all.

------

Oh yea I should have mentioned that I had seen a correct version in an older commit - after a long day of debugging a foreign codebase and porting my Lexer to nearley-lexer I got a little sloppy I guess :)

I am currently building a bigger language with nearley (which is why I am writing PRs making the output nicer 😄 ) - So I will use nearley for the foreseeable future, I would happily review Typescript (and CS / JS for that matter) PRs for you.

Hardmath123:
Thanks for understanding, @alexandertrefz!

> I would argue that people should use a local version…

Ahh, if only everyone did what they should! :-)

> Your solution approach sounds good.

Fantastic! Would you mind helping make the changes? I can make the code-change to the linter if you'd like, but it would be nice if you could at least write the text for me, since I'm not a TS user.

> I think nearleyc should support something like --silent to hide the warning, if you want to show it and you should probably remove it after a couple of versions.

Yes, I completely agree. `-q` and `--quiet` is probably best. I will implement this in the next couple hours.

> I would happily review Typescript … PRs for you.

I appreciate that! I'll be sure to @mention you in TS PRs that confuse me.

alexandertrefz:
I am happy to help, however, here in germany it is now about to be midnight so I'll be signing of for the next few hours, but tomorrow evening I'll jump back in and help out with the linter stuff :)

Hardmath123:
Fantastic! Thank you from here in USA.

tjvr:
Hi @alexandertrefz! We came across a similar issue trying to add ES module support (#299). Does TS support default exports? If so, that would by my preferred way to solve this—probably better than exporting `grammar`, since this is the only object we'll ever want to export from a compiled grammar file. :-)

alexandertrefz:
Hey @tjvr, first of all, sorry for the total lack of progress on this, a bunch of real life and work happened, so I couldnt finish this up.

As for the default export support: Typescript has support for default exports. In general you can find almost any ES Feature in Typescript as it aims to be a strict superset (however, some feature might be lagging behind at any given point in time).

The default export is certainly the cleanest solution, but keep in mind that this will create diverging APIs between ES/TS and CS(At least for Coffeescript 1.x, 2.x has support for exports also, but I don't think support for that has landed).

Also, technically, for Typescript we want to export more than just grammar, we want to export the various interfaces also (I think. It is done now, but im not entirely certain it is necessary, would need to investigate more) so this would mean that all 3 preprocessors entertain differing APIs:

CS: Named exports
ES: Default export of grammar object
TS: Default export of anonymous object containing graamr as well as the interfaces.

That seems undesirable.

If we can get away with not exporting the interfaces and move the CoffeeScript support up to 2.x, we could unify properly with a single default export.

alexandertrefz:
How should we proceed? I am more than happy to align this branch and finish the linter messages (especially since I actually have the time to do it now), just have to tell me which way you guys want to go :)

tjvr:
If possible I think it’s best just to default export the grammar, so one can pass it straight to fromCompiled.

I don’t know if it’s important for TS to be able to see interfaces; I thought interfaces were provided by the `@types/nearley` package? Whatever you think is best. :-)

It might make sense for ES and TS to have the same API (TS is just ES plus types, after all), but I’m not concerned about the CS API diverging: I’m not sure if anyone still uses CS with nearley anyhow!

alexandertrefz:
You are right, `@types/nearley` contains all those interfaces, so we don't need to export them ourselves (We probably can even omit them entirely from that file, but we might then have a dependency towards `@types/nearley`, will investigate).

If the CS is no concern, we can move to a default export. I will adapt the PR. Should I just change TS, or both TS and ES?

tjvr:
Just TS, I think. :) ES modules already has its own PR: #299

alexandertrefz:
Ok, I adapted the PR.

I kept the interfaces in the generated file, to prevent compiler warnings (post-processor functions would have a lot of 'implicit any' warnings otherwise for example) for the end user if they compile TS with `--strict` which is the recommended way. 

I have also looked at the `@types/nearley` types and have found a small missing method (`Lexer.has`) so i'll open a PR for that on the types repo also. We should keep in mind when we change our API that those types need to also be updated.

alexandertrefz:
@tjvr @Hardmath123  Can one of you review this? :)

alexandertrefz:
@kach @tjvr is anything blocking this? Can I help out?

alexandertrefz:
@kach can this be merged?

alexandertrefz:
@kach Pinging again.

alexandertrefz:
@kach I updated this branch again - it would still solve the issues mentioned and has no merge conflicts. Can this be merged? Does it need changing?

kach:
Err, sorry, it's been a while, could you remind me what the status of the PR was? Is this still a breaking change? And, can you update the relevant docs?

alexandertrefz:
Sure thing :)

The PR is basically ready to merge, it does exactly what is described in the description of the PR at the top, still, with the only difference being that we changed to a default export, like i described in this comment: https://github.com/kach/nearley/pull/287#issuecomment-325226542

It is thus, still a breaking change, technically, albeit a very minor one - it is a literal 1 line change, from 

```ts
import * as grammar from './grammar'
import { Parser, Grammar } from 'nearley'
let parser = new Parser(Grammar.fromCompiled(grammar))
```

to

```ts
import grammar from './grammar'
import { Parser, Grammar } from 'nearley'
let parser = new Parser(Grammar.fromCompiled(grammar))
```

but it also enables destructuring of the grammar object if desired:

```ts
import { ParserRules } from './grammar'
```

I am happy to update the docs, if you could point me to the place you think is best.
The same goes for the linter message you wanted to show - if you give me a hint where to change it, ill update the PR.

You should probably be the one to implement the `--quiet` option on the linter, that you talked about here: https://github.com/kach/nearley/pull/287#issuecomment-325226993

alexandertrefz:
Ok, so the exact error message will slightly depend on the individual code. My looks like this for example:


```
error TS2345: Argument of type 'typeof import("/Users/avorin/Projects/essence/essence/src/parser/grammar")' is not assignable to parameter of type 'CompiledRules'.
  Type 'typeof import("/Users/avorin/Projects/essence/essence/src/parser/grammar")' is missing the following properties from type 'CompiledRules': ParserStart, ParserRules
```

for Code that looks like this:

```
import * as grammar from "./grammar"
.
.
.
let parser = new Parser(Grammar.fromCompiled(grammar))
```

The error gets thrown on the line that instantiates the new Parser with a new Grammar, which failes as the object doesnt have the right props.

Unfortunately the previous import statement is a `* as <identifier> from <filename>` import which Typescript resolves to a type of `typeof import(<filename>)` rather than the actual type, which in this case would be: `{ default: CompiledRules }` - if they did otherwise the error would be pretty self explanatory.

Anyways, It seems to me like we should focus on the last part of the second line of the error message: `is missing the following properties from type 'CompiledRules': ParserStart, ParserRules`
Which will stay the same across all variants of this error, and should be relatively unique to this type of issue, where we can then tell the solution to change the import line.

What do you think?

alexandertrefz:
I've added a piece of text to the docs - let me know if you want it changed in anyway :)

alexandertrefz:
I updated the PR.

kach:
Excellent! Merged. Will publish a new minor version on npm this evening.

It only took two years and a day to get this done… thank you again for your patience. :-)

alexandertrefz:
🎉 Thank you for merging it! Glad it is finally in :) 🎉

kach:
Waitwaitwait one other thing — did you update all calls to serializeRules to properly specify that third argument you added?

kach:
Oh, I see, it's optional. Got it.

alexandertrefz:
@kach dont forget about the new NPM Release you wanted to cut ;)

kach:
Yes, done.

Issue #290, opened 2017-09-04
Macro examples in documentation do not compile


fauxneticien:
I wish I could be more helpful than reporting that the macro examples in the documentation do not compile (neither in the nearley playground nor with `nearleyc` 2.11.0 in the Terminal), but I'm only just trying to learn about macros in nearley.

For some reason the compiler is saying `Error: Unknown macro: ____`, even though it is clearly defined in the grammar.

#### In the playground

![screen shot 2017-09-04 at 4 46 06 pm](https://user-images.githubusercontent.com/9938298/30014831-a9a37608-9191-11e7-8ced-bd468f071093.png)
![screen shot 2017-09-04 at 4 46 51 pm](https://user-images.githubusercontent.com/9938298/30014861-bbdf8c80-9191-11e7-9738-2796d22c2404.png)

#### With `nearleyc` 2.11.0

[![asciicast](https://asciinema.org/a/rB8D25j3nteSbuQFIkvbU8fHj.png)](https://asciinema.org/a/rB8D25j3nteSbuQFIkvbU8fHj)

Hardmath123:
Hey @fauxneticien — thanks for the report! At some point, I must have absentmindedly rewritten those examples and forgotten to check if they still compile.

The issue is that macros must be defined before usage. This lets you shadow them if needed, and enforces some sanity when you have multiple `@import`ed grammars (I learned this the hard way!). So, moving the `main ->` lines in the examples to the bottom of the file lets them compile just fine.

```
$ cat - | pbcopy
# Matches "'Hello?' 'Hello?' 'Hello?'"
matchThree[X] -> $X " " $X " " $X
inQuotes[X] -> "'" $X "'"

main -> matchThree[inQuotes["Hello?"]]

$ nearley-test -q -i "'Hello?' 'Hello?' 'Hello?'" <(nearleyc <(pbpaste))
[ [ [ [ [ '\'', [ 'Hello?' ], '\'' ] ],
      ' ',
      [ [ '\'', [ 'Hello?' ], '\'' ] ],
      ' ',
      [ [ '\'', [ 'Hello?' ], '\'' ] ] ] ] ]
```

I fixed this in the docs via d29b5a1. Sorry about that. Maybe we need tests for our docs, now! Meanwhile, please let me know if you have any other issues.

P.S. I just realized how clever your username was. :-)

PR #291, opened 2017-09-05
Split up tests

tjvr:
This:

* Adds some tests for macros (ref #290)
* Tries to split up the tests into tests for `nearleyc`--the compiler--and tests for the parsing algorithm.

The idea is to make each of the parser and nearleyc easier to refactor on their own.

@Hardmath123 Thoughts? :-)

### DIFF
--- a/test/nearleyc.test.js
+++ b/test/nearleyc.test.js
@@ -10,7 +10,7 @@ function read(filename) {
     return fs.readFileSync(filename, 'utf-8');
 }
 
-function inspect(grammar) {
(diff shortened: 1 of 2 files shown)


Hardmath123:
Looks good to me. Feel free to merge.

P.S. I forgot the pretty-printing thing was a thing. It might be interesting to expose that to users? (Or modify it a bit to get LaTeX/HTML/even PEGjs versions for free.)

tjvr:
Going to merge this as part of #311.

Issue #298, opened 2017-09-28
Output grammar as ES module.


independentgeorge:
The fact that the output is UMD wrapped breaks native `import`s used inside the grammar as they are not in top-level scope.

Is adding an ES preprocessor an option?

tjvr:
Hi! Yes, an option to output as a module makes sense to me.

Unfortunately, I don’t think Node supports modules yet—which would make this hard to test!

Would you like to send a PR? The relevant code is in `generate.js`, and the relevant tests are in `nearleyc.test.js`. :)

independentgeorge:
Sure thing, I'll gladly do it!

Is naming the preprocessors `es` and `ecmascript` ok? It might be confused with the default `javascript`.

PR #299, opened 2017-09-29
Add ES modules preprocessor

independentgeorge:
#298 Here it is, edit the names or whatever you think needs editing.

Importing would look like this
```
import { Lexer, ParserRules, ParserStart } from …
import * as grammar from …
```


tjvr:
Thanks for this! Some quick thoughts:

- The preprocessor should probably be called `module` (or perhaps `esmodule`—what do you think, @Hardmath123?)

- I think we'd like to be able to use it like so:

```js
import * as generated from './my-grammar'
const grammar = nearley.Grammar.fromCompiled(generated)
// ...
```

- On second thought, I'm not actually sure if we can use this :( Surely if you can't import the current UMD grammars, then you can't import Nearley either? I'd have thought you just need to convince your module loader that it's okay to import CommonJS stuff. :-)

independentgeorge:
> On second thought, I'm not actually sure if we can use this :( Surely if you can't import the current UMD grammars, then you can't import Nearley either?

The problem is including a grammar which uses `imports` in its preprocessors. With the UMD wrapper, the import statement is nested in an IIFE, and Rollup will throw an error before it tries transpiling to `require`s.

> I think we'd like to be able to use it like so:
> ```
> import * as generated from './my-grammar'
> const grammar = nearley.Grammar.fromCompiled(generated)
> // ...
> ```

It works this way now.

tjvr:
> The problem is including a grammar which uses imports in its preprocessors.

Oh, yes, thanks. I’d forgotten about that!

---

@hardmath123 Looks like adding this is a good idea. Do you approve? :-)

tjvr:
> It works this way now. I tried it out in a project which uses imports all the way [...] and all is good.

Great! Thanks for your explanation above :-)

Would you be up for adding a little example grammar to `grammar/module/`, and its related test in `nearleyc.test.js`?

independentgeorge:
Since node does not support modules natively, the only way to test is to use some tool to transpile it first, right?

Please correct me if I'm wrong, but aren't the coffescript/typescript tests doing the same thing - nearley's result (`.coffee` or `.ts` file) is converted to javascript for node, and the tests are run on that? No environment supports them directly either.

In fact, the test I included ([`nearleyc.test.js`](https://github.com/independentgeorge/nearley/blob/0ac7b528cb15612169a77ecabe46e7f30960fb42/test/nearleyc.test.js#L21) and [`test/grammars/ecmascript-test.ne`](https://github.com/independentgeorge/nearley/blob/0ac7b528cb15612169a77ecabe46e7f30960fb42/test/grammars/ecmascript-test.ne)) is literally a copy of the typescript test which indeed imports moo (and the syntax is identical, in this case at least).

The transpiling tool I used is babel - I imagined it would be enough to make the sure the commonjs equivalents of `import`/`export` work.

> … and then try to import the grammar from a standalone JS test file?

Again, correct me if I'm wrong, but isn't this already tested for when [requiring the transpiled grammar](https://github.com/independentgeorge/nearley/blob/0ac7b528cb15612169a77ecabe46e7f30960fb42/test/nearleyc.test.js#L25)?

lib/generate.js
```diff
@@ -135,6 +135,18 @@ generate.javascript.builtinPostprocessors = {
     "id": "id"
 }
 
+generate.es = generate.ecmascript = function (parser, exportName) {
```
tjvr: Can this be called `module` or `esmodule` please? :)

lib/generate.js
```diff
@@ -135,6 +135,18 @@ generate.javascript.builtinPostprocessors = {
...
+    var output = "// Generated automatically by nearley, version " + parser.version + "\n";
+    output +=  "// http://github.com/Hardmath123/nearley\n";
+    output += "function id(x) { return x[0]; }\n";
+    output += parser.body.join('\n');
+    output += "let Lexer = " + parser.config.lexer + ";\n";
+    output += "let ParserRules = " + serializeRules(parser.rules, generate.javascript.builtinPostprocessors) + ";\n";
+    output += "let ParserStart = " + JSON.stringify(parser.start) + ";\n";
+    output += "export { Lexer, ParserRules, ParserStart };\n";
```
tjvr: I think we want to use `export default` here; importing these objects individually shouldn't be allowed.

tjvr:
> correct me if I'm wrong, but isn't this already tested for here?

Oops! I'm very sorry, I didn't see that :-) [Didn't wait long enough for the complete diff to load...] Great, thank you!

---

The second part is very similar to #304 actually (cc'ing @glmdgrielson):

We already have an existing test for the TypeScript **preprocessor**, and @independentgeorge you've added a test for the esmodules preprocessor here.

It'd be nice if we could also exercise the parser itself: perhaps we could add an ES6 file (and in #304, a TypeScript file) that imports the grammar and tries to use it. The file might look something like:

```js
import * as compiledGrammar from './ecmascript-test.js'
const grammar = nearley.Grammar.fromCompiled(compiledGrammar)
const parser = new Parser(grammar)
parser.feed('foo')
// ... make sure to raise an error, so formatError() gets invoked, to test #304...
```

By compiling the file (with babel/tsc as appropriate), and executing it from `parser.test.js`, perhaps we could test that we can successfully **import** the generated grammar?

Then again, maybe this wouldn't work, and is out of scope of these PRs -- let me know if you think I'm being silly :)

kasbah:
>By compiling the file (with babel/tsc as appropriate), and executing it from parser.test.js, perhaps we could test that we can successfully import the generated grammar?

Once #308 is merged the `sh` function returns `stderr` so we can at least check that after running `coffee` `tsc` `babel` etc.

lib/generate.js
independentgeorge: Are you sure about that? Named exports (the "individual objects") are generally considered best practice. Btw,

    export { Lexer, ParserRules, ParserStart }

is short for

    export { Lexer }
    export { ParserRules }
    export { ParserStart  }

and can be used just like you stated in [this comment](https://github.com/Hardmath123/nearley/pull/299#issuecomment-333079904):

    import * as generated from "./path"
    // generated.Lexer
    // generated.ParserRules
    // generated.ParserStart

Or alternatively:

    import { Lexer, ParserRules, ParserStart } from "./path"
    // Lexer
    // ParserRules
    // ParserStart

independentgeorge: Oh, and you probably want these to be consistent with TypeScript imports/exports (the syntax seems identical to me at first glance), which indeed use named exports and not default.

tjvr: The point is that we want to encourage use of `nearley.Grammar.fromCompiled`. If we were going for named exports, then I'd want it to be a single object (perhaps called "grammar"). But since the only object a generated grammar will _ever_ contain is the grammar, I think we should use a default export please. :-)

independentgeorge: Sure, I understand your reasoning, but it feels very wrong to use one approach for typescript and another for native imports/exports (since they both follow the same spec).

And changing ts to defaults would break existing code... Does that concern make sense?

independentgeorge: @tjvr Since this wasn't answered, I included both named and default exports.

independentgeorge: @tjvr Since you didn't respond to this, I included both default and named exports.

tjvr: Just the one default export please!

It’s true we don’t want to break backwards compatibility with TS, but we can always change TS later (perhaps in v3).

independentgeorge:
> By compiling the file (with babel/tsc as appropriate), and executing it from parser.test.js, perhaps we could test that we can successfully import the generated grammar?
> 
> Then again, maybe this wouldn't work, and is out of scope of these PRs -- let me know if you think I'm being silly :)

Of course, it turns out I was being silly :)

Could you [take a look](https://github.com/independentgeorge/nearley/blob/0d0faaa8cc98519f566a129baa4a706156c19675/test/nearleyc.test.js#L36) and share your thoughts?

independentgeorge:
> By compiling the file (with babel/tsc as appropriate), and executing it from parser.test.js, perhaps we could test that we can successfully import the generated grammar?
> 
> Then again, maybe this wouldn't work, and is out of scope of these PRs -- let me know if you think I'm being silly :)

Of course, it turns out I was being silly :)

Could you [take a look](https://github.com/independentgeorge/nearley/blob/0d0faaa8cc98519f566a129baa4a706156c19675/test/nearleyc.test.js#L36) and share your thoughts?

independentgeorge:
> By compiling the file (with babel/tsc as appropriate), and executing it from parser.test.js, perhaps we could test that we can successfully import the generated grammar?
> 
> Then again, maybe this wouldn't work, and is out of scope of these PRs -- let me know if you think I'm being silly :)

Of course, it turns out I was being silly :)

Could you [take a look](https://github.com/independentgeorge/nearley/blob/0d0faaa8cc98519f566a129baa4a706156c19675/test/nearleyc.test.js#L36) and share your thoughts?

independentgeorge:
> ...however, I can't convince babel to compile the imported grammar (esmodules-test.js) as a module :( Maybe you'll have more luck than I?

Yeah, `externalNearleyc` gives the grammar some random sufix, so we can't reallly import it statically unless renamed, or simply create the module on the fly:

```javascript
it('builds for ES6+', function() {
    this.timeout(10000);

    // Compile the grammar
    const {outPath, stdout, stderr} = externalNearleyc("grammars/esmodules-test.ne", '.js');
    expect(stderr).toBe("");
    expect(stdout).toBe("");

    // Create the re-export/parse test module
    write(`test/${outPath}-parse.js`, `import {Grammar, Parser} from '../lib/nearley'
                                       import compiledGrammar from './${outPath}'

                                       const grammar = Grammar.fromCompiled(compiledGrammar)
                                       const parser = new Parser(grammar)
                                       parser.feed('<4>')
                                       console.log(parser.results)`);

    // Notice babel-node, which compiles modules prior to running them
    { const {stderr, stdout} = sh(`babel-node ${outPath}-parse.js --presets=env`);
    expect(stderr).toBe("");
    expect(stdout.trim()).toBe("[ [ '<', '4', '>' ] ]"); }
});
```

The `write` part could be extracted to a function, making it possible to feed the parser multiple inputs (since they too would be hardcoded in the module).

independentgeorge:
Any news on this?

tjvr:
I think this looks good! Thanks for working on this :-)

cmcaine:
> I was using the wrong terminology before. We want to use a default export, and that should be the only way to use this:

This means that grammars generated by the typescript and module preprocessors have to be imported in different ways. Perhaps both outputs should support both types of import?

Non-default imports also allow easier tree-shaking, if one doesn't actually need every part.

cmcaine:
And +1 for docs. The preprocessors aren't mentioned anywhere and it's taken me an embarrassingly long time to work out how to use nearley with typescript.

Here's a little demo repo for typescript with nearley: https://github.com/cmcaine/nearley-ts-demo

It's not hard once you know, of course.

tjvr:
> And +1 for docs

Pull requests are welcome! I don't have the bandwidth to get to this right now, and TS support is pretty low on my list unfortunately :)

independentgeorge:
Well, they are [briefly mentioned](https://nearley.js.org/docs/parser#using-preprocessors), but the paragraph can easily be overlooked. I suppose an example or two would help.

Issue #302, opened 2017-09-30
Typing issue


glmdgrielson:
As a fun little thing I tried porting this to Typescript and I found a big hole in the Typescript generation, namely StreamLexer doesn't match the Lexer interface you provided. Oops...
```
Class 'StreamLexer' incorrectly implements interface 'Lexer'.
  Types of property 'formatError' are incompatible.
    Type '(token: any, message: any) => any' is not assignable to type '(token: Token) => string'.
```

tjvr:
We’ve had some problems recently with the typescript bindings; they’re not written or maintained by the core maintainers (@hardmath123 and I).

Is this a bug in the generator, or in the typescript bindings? Thanks.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

glmdgrielson:
I think it's in the bindings:
```typescript
export interface Lexer {reset:(chunk:string, info:any) => void; next:() => Token | undefined; save:() => any; formatError:(token:Token) => string; has:(tokenType:string) => boolean};
```
The error I got came when I made StreamLexer implement Lexer. Also, about the TypeScript stuff: If you need help with that, I'd be willing to lend a hand!

tjvr:
Sorry. By the typescript bindings, I meant the `@types/nearley` package: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/nearley

If you’ve found a bug in `generate.js`, feel free to send a PR :-)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

glmdgrielson:
Okay, yeah it's definitely a bug in `generate.js`. I'll try to send a PR soon, because both the default lexer AND the recommended one - yours, [moo](https://github.com/no-context/moo) - fail to match the type  specified by the Lexer definition.
What you have is: 
```typescript
formatError(token: Token, message?: string): string;
```
What you require is: 
```typescript
formatError: (token: Token) => string;
```

glmdgrielson:
The pull request is up.

PR #304, opened 2017-10-01
Fix Typescript template

glmdgrielson:
Fixes the Lexer typing for 'formatError' to agree with its actual use.

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -179,7 +179,7 @@ generate.ts = generate.typescript = function (parser, exportName) {
     output += "  reset: (chunk: string, info: any) => void;\n";
     output += "  next: () => Token | undefined;\n";
     output += "  save: () => any;\n";
-    output += "  formatError: (token: Token) => string;\n";
+    output += "  formatError: (token: Token, message?: string) => string;\n";
     output += "  has: (tokenType: string) => boolean\n";
     output += "};\n"
     output += "\n";


tjvr:
Note to self: This is for #302.

tjvr:
Hey, thanks for sending this :-) Your change looks right to me (although I don't use TS myself!).

We'd like to update the tests, to make sure this doesn't break again! Perhaps we want to add a `.ts` file to `test/grammars/`. Then we could add a test suite to `test/nearleyc.test.js` which, compiles `typescript-test.ne` like the existing test does, but then runs `tsc` on our new file, to check TypeScript is happy with using the parsing API. Would you be up for doing this? :D

glmdgrielson:
Yeah, sure! Any suggestions for such an example?

glmdgrielson:
#299 could probably be solved by renaming the Typescript test! So I'm adding another file `es6-test.ne` and re-doing an example. Also, can I assume that the data array will contain the output of the postprocessors?

independentgeorge:
> #299 could probably be solved by renaming the Typescript test!

Could you please explain? I can't see the connection of that with typescript.

glmdgrielson:
None of the syntax there is exlusive to Typescript. There's no typing info or such; the only reason it uses Typescript is for ease of use with ES6 style imports.

tjvr:
It's true (and amusing :P) that in this case, the file might be both valid TS and valid ES6; but we'd rather have separate script files for each test, please :-)

glmdgrielson:
Hey I've made a few changes. I think the tests should work. Crontab was _always_ giving me issues so I don't know what to make of it.

tjvr:
Hi @glmdgrielson! Thanks for working on this :)

Crontab often gives me issues too, I'm not sure what's up with that test. I wouldn't worry about it.

Thanks for adding a test grammar, that's cool! Unfortunately the new test doesn't seem to break, if I undo your original fix :/ Do you think it's possible to make it so the test checks `formatError` type-checks properly? Or should I merge this as-is? :-)

test/nearleyc.test.js
```diff
@@ -60,15 +65,27 @@ describe("bin/nearleyc", function() {
...
 
     it('builds for TypeScript', function() {
-        this.timeout(10000); // It takes a while to run tsc!
+        this.timeout(100000); // It takes a while to run tsc!
         const {outPath, stdout, stderr} = externalNearleyc("grammars/typescript-test.ne", ".ts");
         expect(stderr).toBe("");
         expect(stdout).toBe("");
-        sh(`tsc ${outPath}.ts`);
+        sh(`tsc ${outPath}.ts --noEmitOnError`);
+        expect(stderr).toBe("");
```
tjvr: I don't think this will do what you think it does: `stderr` is still the output from `nearleyc` :)

glmdgrielson:
To answer your question, just merge. I don't really know of a way to make it type-check... `tsc` erases it all by compile time anyway.

Issue #230, opened 2017-04-30
@include in multiple files


kasbah:
At the moment you can't seem to be able to include a file in multiple files that are themselves also included.

grammar.ne
```
@include "a.ne"
@include "c.ne"

main -> a | c
```

a.ne
```
@include "b.ne"

a -> "a" | b
```

b.ne
```
b -> "b"
```

c.ne
```
@include "b.ne"

c -> "c" | b
```

Results in a warning: 
```
WARN	Undefined symbol `b` used.
```

If c.ne is excluded then there is no warning. It severely limits you in using several files and re-using logic.

somebody1234:
No, it works perfectly fine, the issue is with the linter.

kasbah:
Another related issue is that nearleyc get's confused about the directory:
```
.
├── dir
│   ├── a.ne
│   └── b.ne
└── grammar.ne
```

grammar.ne
```
@include "dir/a.ne"

main -> a
```
a.ne
```
@include "b.ne"

a -> "a" | b
```
b.ne
```
b -> "b"
```


```
$ nearleyc grammar.ne -o grammar.js
fs.js:641
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/tmp/eg/b.ne'
    at Error (native)
    at Object.fs.openSync (fs.js:641:18)
    at Object.fs.readFileSync (fs.js:509:33)
    at Compile (/tmp/eg/node_modules/nearley/lib/compile.js:49:39)
    at Compile (/tmp/eg/node_modules/nearley/lib/compile.js:53:25)
    at StreamWrapper.<anonymous> (/tmp/eg/node_modules/nearley/bin/nearleyc.js:57:17)
    at emitNone (events.js:91:20)
    at StreamWrapper.emit (events.js:185:7)
    at finishMaybe (_stream_writable.js:512:14)
    at endWritable (_stream_writable.js:522:3)
```

PR #308, opened 2017-10-05
fix lint of multi includes of same file

kasbah:
Looking at #230, starting with a test.


lib/compile.js
```diff
@@ -61,6 +60,7 @@ function Compile(structure, opts) {
                 Object.keys(c.macros).forEach(function(k) {
                     result.macros[k] = c.macros[k];
                 });
+                require('./lint.js')(result, {out: process.stderr});
```
kasbah: Even though this fixes the issue, I think it still breaks `--quiet` option. Does it need to lint inside compile at all?

test/lint.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+    it("doesn't warn about duplicate symbol", function () {
+        mockGrammar.rules =  [
+            {name: "a", symbols: []},
+            {name: "a", symbols: []},
+            {name: "b", symbols: ["a"]}
+        ];
+        lint(mockGrammar, mockOpts);
+        expect(writeSpy).toNotHaveBeenCalled();
+    });
```
kasbah: So, is that as expected? Would be better if the linter warned about duplicate rules right?

test/lint.test.js
```diff
@@ -0,0 +1,52 @@
...
+    it("doesn't warn about defined symbol", function () {
+        mockGrammar.rules =  [
+            {name: "a", symbols: []},
+            {name: "b", symbols: ["a"]}
+        ];
+        lint(mockGrammar, mockOpts);
+        expect(writeSpy).toNotHaveBeenCalled();
+    });
+
+    it("doesn't warn about duplicate symbol", function () {
```
kasbah: So, is that as expected? Would be better if the linter warned about duplicate rules right?

lib/compile.js
tjvr: Hmm. It looks like `compile.js` only does this in order to lint `@include`-d files.

Maybe we should rewrite `lint.js` to recursively process included files by itself; I don't think we should be requiring `lint` from inside `compile`.

tjvr: ...`lint` appears to only run on the generated grammar. Now I'm _very_ confused!

test/lint.test.js
tjvr: No, this is definitely expected: we want to allow multiple rules for same non-terminal :-)

tjvr: Oh, you mean _duplicate_ rules, as in, with the same name and symbols. That might be a nice thing to test for, but I think it's outside the scope of this PR :/

lib/compile.js
kasbah: What do you mean by that last statement? As it stands lint is run in `lib/compile.js` every time a new file is included and then [in nearleyc](https://github.com/Hardmath123/nearley/blob/aa77dee0d1fc728a5ffad046d4baed9eab49eaed/bin/nearleyc.js#L61).

test/lint.test.js
kasbah: Well, I only half know what I am doing here. [Lark handles the new multi-include](https://github.com/erezsh/lark/pull/28#issuecomment-334559868) test very differently and complains loudly about re-declared things, is that related? 

How would you change this test  for the actual desired behavior? Add a symbol to the second a rule?

```diff
             {name: "a", symbols: []},
-            {name: "a", symbols: []},
+            {name: "a", symbols: ["b"]},
             {name: "b", symbols: ["a"]}
```

lib/compile.js
kasbah: Do you mean it _should_ only be run on generated grammars but isn't or it should _not_ be run on generated grammars but is?

test/nearleyc.test.js
```diff
@@ -14,46 +14,46 @@ describe("bin/nearleyc", function() {
     after(cleanup)
 
     it('builds for ES5', function() {
-        const out = externalNearleyc("grammars/parens.ne", '.js')
-        var grammar = nearley.Grammar.fromCompiled(require(`./${out}.js`));
+        const out = externalNearleyc("grammars/parens.ne", '.js');
+        const grammar = nearley.Grammar.fromCompiled(require(`./${out}.js`));
```
tjvr: Not a big deal here, but please avoid doing style changes in future PRs, without asking first; they make things harder to review. Thanks! :-)

kasbah: Yes, sorry, it was annoying me and I couldn't resist. I can remove the commit from this PR if you want.

tjvr: Nah, not a big deal! I was going to get round to it eventually... :-)

kasbah:
Thanks for reviewing @tjvr tjvr, 

> Apologies for my earlier comments about require('./lint'); I've somewhat confused myself trying to figure out what's going on here. I propose two things:
> 
>- Please can you add a test for the original issue? ("fix lint of multi includes of the same file"). I can't find one.

It is actually added in [the first commit](https://github.com/Hardmath123/nearley/pull/308/commits/0c8c3ba476a9f5f4d3f274a98ef57071ee7e3602). Before, `externalNearleyc` was seemingly capturing `stderr` and making sure it was empty but was _actually_ capturing `stdout`. Now it captures both and makes sure they are empty. 


>- Can you explain why [this change](https://github.com/Hardmath123/nearley/pull/308/files#diff-71a1f6779f4a9a1fe69fcec81f85db12R63) is necessary? I can't work out why you've moved
> 

Instead of only linting the grammar from the current file (variable `c`) it now lints the accumulated grammar (variable `result`) so it won't emit a warning in the new multi-include test case. 

> Then, I think I'm going to merge this, and have a look at refactoring compile.js a little bit; I'm not sure it should be requiring lint.js at all!

Well, it wouldn't really be a refactor as it would change current behaviour, more of a bug fix. :beetle: 

The only reason I can see why it's there is so that it will print warnings to stderr when `Compile` is used from node without using nearleyc. Like I said before, it does mean the `--quiet` option is broken so I would consider it a bug. I don't know if I would break anything by just removing the lint at that point. Looks like it was introduced with 44e1b693e4c5d961dcae8ea5c9a26c0a172fc49f but that doesn't really help. Maybe @Hardmath123 can shed some light.

test/external.js
```diff
@@ -8,14 +8,15 @@ const expect = require('expect');
 const root = 'test/'
 
 function sh(cmd) {
-    return child_process.execSync(cmd, {encoding: 'utf-8', stdio: 'pipe', cwd: root});
+    return child_process.spawnSync(cmd, {shell: true, encoding: 'utf-8', stdio: 'pipe', cwd: root})
```
tjvr: What's the reason for using `spawn` instead of `exec`? :-)

test/external.js
```diff
@@ -8,14 +8,15 @@ const expect = require('expect');
...
 }
 
 var highestId = 0
 function externalNearleyc(input, ext) {
     const tempPath = 'tmp.' + path.basename(input) + (++highestId)
-    const stderr = sh("../bin/nearleyc.js " + input + " -o " + tempPath + ext);
+    const {stderr, stdout} = sh("../bin/nearleyc.js " + input + " -o " + tempPath + ext);
     expect(stderr).toBe("")
+    expect(stdout).toBe("")
```
tjvr: Nice!

test/external.js
kasbah: So we can get back `stderr`. The docs for child_process are a hard to read so I may have missed something, but this definitely works.

tjvr:
Thanks for your explanation! :)

> it does mean the --quiet option is broken so I would consider it a bug

I agree, this is a bug. Can we add a test for the `--quiet` option please? :D

I imagine the `require('lint')` line was added to make sure @included files get linted too. But given that `nearleyc` runs the linter over the whole grammar anyway, I'm not sure I understand why we need this line in `compile.js` at all? Perhaps we should remove it in this PR.

kasbah:
Remove it and see who complains? :smiling_imp:

test/lint.test.js
kasbah: @tjvr could you give some feedback on this please

tjvr: I suggest leaving this as-is. The current behaviour is expected.

tjvr:
Sure, go for it! :)

Please could you add a test for the --quiet option? Then I think we should merge this :)

kasbah:
Ace, should be good to go then.

test/external.js
tjvr: Okay, great. :)

PR #311, opened 2017-10-07
add test for macros & split up tests

kasbah:


tjvr:
Thanks for this! I actually tried to add some macro tests in #291, I've been meaning to merge that for a while. (I was waiting to double-check it.)

I would merge it now, but sadly it now conflicts--I think because of your style changes :-(

If you wouldn't mind extending my PR to integrate your macro tests, and also deconflicting it, that'd be grand! :-)

kasbah:
I merged your branch in. I was afraid duplicate tests would appear but I can't find any.

tjvr:
Thanks for deconflicting! I removed the duplicate tests.

test/nearleyc.test.js
```diff
@@ -207,3 +176,48 @@ describe('builtins', () => {
...
+    it('must be defined before use', () => {
+        expect(() => compile(`
+            main -> matchThree[inQuotes["Hello?"]]
+            matchThree[X] -> $X " " $X " " $X
+            inQuotes[X] -> "'" $X "'"
+        `)).toThrow();
+    });
+
+    it('compiles a simple macro from external file', function() {
+        const grammar = compile(read("test/grammars/macro-test.ne"));
```
tjvr: Please could we use a template string `\`...\`` instead of `read()`? I'd rather bring the grammars into the test files themselves where possible (i.e. they're not too long).

kasbah: Well, part of the reason I added the test was actually to be able to reuse it for Lark and it needs to be an external file for that. https://github.com/erezsh/lark/issues/26

tjvr: Huh, fair enough.

Issue #310, opened 2017-10-06
Add a linter or formatter


kasbah:
Add a linter and/or formatter. Something that requires zero configuration and just fixes style issues with a single invocation. 

Maybe one of: 
- [prettier](https://github.com/prettier/prettier)
- [standard](https://standardjs.com/)

Or even: 
- [prettier-standard](https://github.com/sheerun/prettier-standard)

Any preferences?

tjvr:
I’m not generally a fan of linters; they tend to disagree with how I want my code to look :-)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kasbah:
In that case maybe you'd prefer to set up an eslint config that describes the current preferred style. The advantage of the ones I mentioned is that you don't have to spend the time doing that and you just drop them in and they do their thing. 

In general I would encourage you to get over the "how you want your code to look" as you always need to conform to the existing style when working on other people's code. I say: if it's consistent then be happy; anything else is a waste of time. 

Right now this code base, though it isn't too bad at all, is not consistent because there is no formatter in place.

tjvr:
Yeah, it would be lovely if the codebase were more consistent :-)

I actually hadn't come across `prettier` before; I've just been experimenting with `prettier--no-semi --trailing-comma es5` for [Moo](https://github.com/tjvr/moo) and it _almost_ does what I want. I like that it operates on the AST, like `gofmt` (and unlike ESLint AFAIK).

However, this is @Hardmath123's project; and of the two of us, he tends to be more resistant to changes like this :-) Perhaps we could experiment with adding it to the tests, and perhaps we can come around to adding it to the core library later. How does that sound? Would you like to send a PR? (With the caveat that @Hardmath123 might reject it! :P)

PR #313, opened 2017-10-09
Add code formatter

kasbah:
As discussed in #310

This adds an npm script: `npm fix`, that will reformat the code using prettier
and a bit of eslint (mainly to fix use of `const` and `=>` in tests). Eslint errors that can not be automatically fixed are currently printed but otherwise ignored (exit code is 0).

If you think it's a good idea it's best to merge all pending pull-requests first at which point I can rebase this and it can be merged and we will have more consistent code going forward.

### DIFF
--- a/bin/.eslintrc
+++ b/bin/.eslintrc
@@ -1,0 +1,14 @@
+{
+  "env": {
+    "node": true,
+    "commonjs": true,
+    "es6": false
+  },
+  "rules": {
+    "indent": [
+      "error",
+      4
+    ]
+  },
+  "extends": ["eslint:recommended"]
+}
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
     "test": "mocha test/*.test.js",
     "doctoc": "doctoc --notitle README.md",
     "profile": "bin/nearleyc.js test/grammars/parens.ne > test/grammars/parens.js && node test/profile.js",
+    "fix": "npm run eslintfix; npm run prettier",
+    "eslintfix": "eslint --fix lib/; eslint --fix test/; eslint --fix bin/",
     "prettier": "prettier --tab-width 4 --write lib/**/*.js test/**/*.js bin/**/*.js"
   },
   "author": "Hardmath123",
@@ -50,6 +52,7 @@
     "benchr": "^3.2.0",
     "coffee-script": "^1.10.0",
     "doctoc": "^1.3.0",
+    "eslint": "^4.8.0",
     "expect": "^1.20.2",
     "microtime": "^2.1.2",
     "mocha": "^2.5.3",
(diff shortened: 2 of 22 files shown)


bin/.eslintrc
```diff
@@ -0,0 +1,14 @@
+{
+  "env": {
+    "node": true,
+    "commonjs": true,
+    "es6": false
+  },
+  "rules": {
```
bd82: eslint indent rules are duplicating what prettier does.
See https://github.com/prettier/prettier#eslint for details how to integrate both these tools at the same time.

package.json
```diff
@@ -34,7 +34,10 @@
     "benchmark": "benchr test/benchmark.js",
     "test": "mocha test/*.test.js",
     "doctoc": "doctoc --notitle README.md",
-    "profile": "bin/nearleyc.js test/grammars/parens.ne > test/grammars/parens.js && node test/profile.js"
+    "profile": "bin/nearleyc.js test/grammars/parens.ne > test/grammars/parens.js && node test/profile.js",
+    "fix": "npm run eslintfix; npm run prettier",
+    "eslintfix": "eslint --fix lib/; eslint --fix test/; eslint --fix bin/",
+    "prettier": "prettier --tab-width 4 --write lib/**/*.js test/**/*.js bin/**/*.js"
```
bd82: use "list-different" option to fail the CI if the formatting deviated from the standard
of prettier.

See: https://github.com/prettier/prettier#--list-different

package.json
```diff
@@ -49,10 +52,12 @@
     "benchr": "^3.2.0",
     "coffee-script": "^1.10.0",
     "doctoc": "^1.3.0",
+    "eslint": "^4.8.0",
     "expect": "^1.20.2",
     "microtime": "^2.1.2",
     "mocha": "^2.5.3",
     "moo": "^0.3.2",
+    "prettier": "^1.7.4",
```
bd82: It is better to use a fixed version number of prettier without a wildcard ('^') because
minor changes may cause formatting changes which will fail the CI build (when using --list-different). 

See: https://github.com/prettier/prettier#usage for details (at end of this section).

kasbah:
Thanks for the review @bd82, I have added you as a collab on my fork so you can make suggested changes to this PR yourself if you want, otherwise I'll take another look once it's decided whether maintainers want this at all. 

As far as I understand it, while your suggestions make sense, the effect on the actual source code would be the same. Maintainers can still look at the existing diff and decide if they want it, right?

bd82:
Yes. the effect on the source code would be the same.
The suggestions would make it easier to maintain the changes over time by the maintainers of Nearley.

bd82:
@kasbah I am sorry but the review is the limit of my collaboration here 😢

kasbah:
No problem, thanks for your suggestions. I'll definitely implement them if it is planned to be merged in general.

tjvr:
Closing in favour of #314. Thanks for looking into this! :)

PR #314, opened 2017-10-10
Code formatter

tjvr:
Previously #313.

* The script is called `fmt`, after `go fmt`.
* I'm still not thrilled about this; it would be nice to tidy things up and make things more consistent, but automated linting tools can never make code as pretty as a human would.
* I split the major changes (re-indenting, double-quoted strings, reformatting) into separate commits, to make the changes easier to review (and make sure we don't break anything!).

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -44,6 +44,7 @@ Before submitting a **pull request**, have you done the following?
   without your change to detect possible bugs, and included the output in your
   pull request as a separate comment. Parsing is tricky, and innocent changes
   can sometimes kill performance. :-(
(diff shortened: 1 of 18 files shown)


tjvr:
@kasbah What do you think? :)

Might need to be rebased/master merged, of course...

kasbah:
I think it's fine though it doesn't catch the `const`  and `=>` stuff that was annoying me in the tests which I did with ES Lint in my PR. Putting anything like this in place is better than nothing though.

tjvr:
> it doesn't catch the const and => stuff that was annoying me in the tests

Yeah, keeping the tests consistent is lower-priority; you're of course welcome to submit PRs to fix the style there yourself! :-)

Issue #318, opened 2017-10-19
multiple-include is broken (with possible fix)


guoyr:
The 'builds modules with multiple includes of the same file' test in `nearleyc.test.js` does not seem to actually work because The `@include "b.ne"` in `c.ne` is not executed, resulting in `b` being undefined in `c.ne`. 

The reason the test passes is that `compile.js`'s `buildToken` function does not check if the token has been previously defined when the token is a string. Changing the token to a macro causes the test to fail.

I have a gist here that shows this bug, with a proposed fix in `compile.js`: https://gist.github.com/guoyr/4615bc3c84b4058e901b93741ea0a306

#230 may be related, but I haven't looked into the fix for that.

tjvr:
Thanks for spotting this!

If you’d like to turn this into a PR, that’d be great :)

PR #319, opened 2017-10-20
Fix multiple includes

guoyr:
This is a PR for the fix proposed in issue #318

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -6,20 +6,24 @@
     }
 }(this, function(nearley) {
 
-    function Compile(structure, opts) {
+    function Compile(structure, opts, prevResult) {
         var unique = uniquer();
         if (!opts.alreadycompiled) {
             opts.alreadycompiled = [];
         }
 
+        if (!prevResult) {
+            prevResult = {};
+        }
+
         var result = {
-            rules: [],
+            rules: prevResult.rules || [],
             body: [], // @directives list
-            customTokens: [], // %tokens
-            config: {}, // @config value
-            macros: {},
+            customTokens: prevResult.customTokens || [], // %tokens
+            config: prevResult.config || {}, // @config value
+            macros: prevResult.macros || {},
             start: '',
-            version: opts.version || 'unknown'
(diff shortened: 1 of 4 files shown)


lib/compile.js
```diff
@@ -50,7 +50,7 @@
                     var parserGrammar = nearley.Grammar.fromCompiled(require('./nearley-language-bootstrapped.js'));
                     var parser = new nearley.Parser(parserGrammar);
                     parser.feed(f);
-                    var c = Compile(parser.results[0], {file: path, __proto__:opts});
+                    var c = Compile(parser.results[0], {file: path, alreadycompiled:[], __proto__:opts});
```
tjvr: I realise this makes your modified test pass, but why does this fix the problem? Won't this make us compile some files more than once?

Also, since you're already changing this line, I propose we use `Object.assign` instead of `__proto__` here. :)

kasbah: Just a note: `Object.assign` requires a polyfill for ES5 environments and everything in `lib` seems to be ES5 so far.

guoyr: Apologies for the late reply.

@tjvr Re your first point: Thanks for catching that issue. I've put up a new patch that will forward previously compiled rules to the recursive call to `Compile()`,  so each file will no longer be compiled more than once.

Re your second point. I had the same thought as Kaspar. But I'd be more than happy to start using  ES2015 going forward, but I think it should at least be in a separate commit that mentions ES2015, so people runner (very) outdated versions of JS engines can know what happened without having to dig through the code.

tjvr:
Why do we need `prevResult` here? I feel like there should be a nicer way of doing this... 🤔

rtimmons:
(for posterity for anyone else running into this and searching for a solution, this change has been incorporated into [this fork](https://github.com/mongodb-forks/nearley) and [published to npm](https://www.npmjs.com/package/@mongodb-js/nearley). If anybody has a better/different way to fix this without the `prevResult` trick then perhaps they can take up the mantle and get this merged. Repeated thanks to the nearley core devs for a great project it's been a real boon to our project.)

PR #330, opened 2017-12-12
Added tokens from lexer as terminals in railroad diagram

MofX:
If you use e.g. moo as a lexer, terminals are shown as "[Unimplemented]" in the railroad diagram. This adds them as terminals with the token name


Hardmath123:
Hey, thanks for the fix! This is a thing that I've been meaning to get to for a while — glad someone could help. :-)

tjvr:
🎉 Hooray!

OOI, what do they look like? :)

Hardmath123:
OOI?

Issue #334, opened 2018-01-20
Is there a public API to generate strings from a compiled grammar?


cemulate:
I would like to generate strings from a compiled grammar `.js` file in the browser -- i.e., exactly what `nearley-unparse` does. However, I can't figure out from the documentation how to accomplish this. The functionality appears to be entirely contained in `nearley-unparse.js`, but said file is not any sort of `import`able module.

It seems to me like the generation functions should be part of a core library that could then be imported in `nearley-unparse.js` or any other arbitrary place in user code. 

Am I missing something?

tjvr:
Closing, since the patch was merged! 🎉

PR #335, opened 2018-01-21
Refactor nearley-unparse generation code into core

cemulate:
I basically moved the generation functions from `nearley-unparse` into the core `nearley` module as a `Generator` class that gets exported along with the other core types.

Now `nearley-unparse` simply acts as a script wrapper around the `Generator` functionality. 

This exposes the generation functionality to arbitrary user code.

I tried to fit the existing code style, and `npm run test` doesn't report any (new) failures.

If merged, this would resolve the concern I raised in issue #334

### DIFF
--- a/bin/nearley-unparse.js
+++ b/bin/nearley-unparse.js
@@ -4,6 +4,7 @@ var fs = require('fs');
 var nearley = require('../lib/nearley.js');
 var nomnom = require('nomnom');
 var randexp = require('randexp');
+var Unparse = require('../lib/unparse.js');
 
 var opts = nomnom
     .script('nearley-unparse')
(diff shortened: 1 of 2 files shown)


Hardmath123:
Nice idea, but I don't want `randexp` to be a dependency of the main nearley.js parser script (which folks include using <script> tags in client-side code).

cemulate:
Ah, yeah; good point.

To be clear, are regex *charsets* the only kind of regex supported in rules? If so, that's sufficiently small in scope that I could whip up a specific generator for charsets and do away with the `randexp` dependency. 

Any other ideas on how to resolve this?

cemulate:
I think another "canonical" way to handle this (i.e., done in other common projects) would be to modulify the whole project (break out `Parser`, `Grammar`, etc. into modules, etc.) and produce a minified distribution build with, say, Webpack. That way, people using `<script>` tags could use the minified dist build, and people using Webpack or other module builders could `import` what they need.

I don't think it would be too hard, and I'd be willing to take that on, but I would want a Yea or Nay from @Hardmath123 before I would start in on it since its a large structure change.

tjvr:
> modulify the whole project

I don't think this is something we're going to do at this time. If we did, I think the main separation would be

* `nearleyc` (the compiler)
* `nearley` (the current `lib/nearley.js`, which contains the parsing algorithm, including `Parser` and `Grammar`)

However, @Hardmath123 has expressed strong reservations about this in the past, so we'd rather not re-visit this discussion :)

cemulate:
@tjvr How about something like the commit I just pushed? Now there's just a `lib/unparse.js` module along with all the rest that `nearley-unparse` `require`s to do its work. 

In my case, since I'm using Webpack I can use it in the browser as well.

Hardmath123:
@tjvr You wanna look this over? lgtm

cemulate:
Hey guys, feel free to merge, but just FYI I was having some serious performance problems when trying to use the unparse routines to do "serious" fake-text generation. I ended up writing [this](https://github.com/cemulate/nearley-generator) as a dedicated project to efficiently generate strings from a compiled Nearly grammar.

tjvr:
I love your site :joy:

Thanks for the link! Unparsing is a reasonably hard problem, particularly dealing with non-termination. Nice to see another approach to it!

kach:
Whoa, neat site! I'm going to put that in the nearley hall of fame.

Tim, the PR still looks good to me. If you think so, too, I'm open to merging it.

cemulate:
Thanks guys! I really appreciate being able to use your library to write & compile the grammar!

tjvr:
LGTM! 🎉

PR #338, opened 2018-01-26
Renamed github corners to use new repo username

RobRoseKnows:
When you changed your username you left the old URL as most of the links. This changes the ones in the github corners by changing it in `template.html` and then recompiling. I don't know whether you'll want to change all the others, I'd be happy to do that if you want.


kach:
I believe old links should redirect to the new one automatically? In any case, this PR seems worthwhile so I will merge it. Thanks.

RobRoseKnows:
They do redirect automatically, but changing them keeps everything consistent. Thanks!

PR #345, opened 2018-02-10
Change postprocessors to nuller function in whitespace.ne

coolreader18:


coolreader18:
nevermind, nuller doesn't work yet i guess

PR #354, opened 2018-03-01
A fix for multiple Parsers operating at the same time with the same grammar containing custom lexers

warnio:
# A small fix for the lexers
I've been working very long and hard on this bug, I would appreciate any efforts for this request. 

## The Code
My problem started here
#### ` js/parser.ne`
```ne
@preprocessor typescript

@{%
import * as Nearley from 'nearley';

@lexer lexer

function parse(type: string, text: string) {
    console.log("NESTED PARSE")
    let grammar = Object.assign({}, require("./grammar"))
    grammar.ParserStart = type;
    // Here it already goes wrong
    const parser = new Nearley.Parser(Nearley.Grammar.fromCompiled(grammar))
    try {
        parser.feed(text)
        return parser.results[0]
    } catch (e) {
        return null;
    }
}
%}
```

## Explanation
In this code I define a function in a Nearley file, which parses a given text using a moo lexer with the given type, and returns the first result. 

## The problem
The problem is that when the parser object is created, it creates a reference to the lexer. This results in a malfunction in the lexer you can see here:

#### Lines changed in `nearley.js`
```
Parser.prototype.feed = function(chunk) {
    var lexer = this.lexer;
    console.log("BEFORE: " + lexer.buffer)
    lexer.reset(chunk, this.lexerState);
    console.log("AFTER: " + lexer.buffer)

    var token;
    // ...
}
```

#### Executed in console
```
var parser = new Nearley.Parser(Nearley.Grammar.fromCompiled(grammar))
parser.feed("/print Hello true world").results
// see grammar down
```

#### Output
```
BEFORE: 
AFTER: /print Hello true world
NESTED PARSE
BEFORE: /print Hello true world
AFTER: true
[["command",[["word","print"],["whitespace"," "],["word","Hello"],["whitespace"," "],["value","boolean",true]]]]
```

As you can see, the lexers from the two parsers are references, so when the buffer of one changes, the buffer of the other one also changes, resulting in a failing `lexer.next()`, which stops the parser from parsing the rest

## What I did next
After **two weeks** of searching through the library, I found a fix to let multiple Parsers containing custom/moo lexers operate at the same time with the same grammar, by changing a few lines. Then I spent **2h+** on making this pull request.

## How I fixed it
First I tried to use `Object.assign({}, grammar.lexer)`, but that didn't work because the resulting object was recognized as a `Object`, and not as a `Lexer`. Soon I found out about the `Lexer.clone()` method from Moo, which did exactly what it was supposed to do: to break the reference. This will only work if the `synchronous` option, which allows clones to be made, is set to `true`.

## Effects
The only thing that needs to happen with this minor change, is that the tokenizer will need to have an extra `clone()` method, which in a normal lexer wouldn't need much effort to make.  

## Benchmarks
The benchmarks here are pretty similar.

#### Without fix
```
$ npm run benchmark

> nearley@2.11.1 benchmark /Users/dylan/Developer/Projects/Nearley
> benchr test/benchmark.js

• test/benchmark.js:
  • calculator.
    ✔  nearley  9,071.75  ops/sec  ±2.18%  (82 runs)
  • json.
    ✔  nearley  77.23  ops/sec  ±1.78%  (66 runs)
  • tosh.
    ✔  nearley  1,397.43  ops/sec  ±3.70%  (81 runs)
```

#### With fix
```
$ npm run benchmark

> nearley@2.11.1 benchmark /Users/dylan/Developer/Projects/NearleyFix
> benchr test/benchmark.js

• test/benchmark.js:
  • calculator.
    ✔  nearley  9,114.15  ops/sec  ±1.30%  (81 runs)
  • json.
    ✔  nearley  77.30  ops/sec  ±2.19%  (67 runs)
  • tosh.
    ✔  nearley  1,480.90  ops/sec  ±1.48%  (86 runs)
```

## @Types
With this request `ParserOptions` in `index.d.ts` needs to be changed to this:

```
export interface ParserOptions {
    synchronous?: boolean;
    keepHistory?: boolean;
    lexer?: Lexer;
}
```

## More Code
Here is some more of my code. Though it won't work, it's supposed to give you an idea. If you have any comments on it, feel free to let me know. 

#### `game/command.ne`
```ne
@preprocessor typescript

@include "../js/parser.ne"

# A lot of includes
@include "../util/text.ne"
@include "../util/whitespace.ne"
# ...
@include "../const/value.ne"

@lexer lexer

command -> "/" any {% (d: any[]) => {
    const strings = d[1]

    let params: any[] = []
    for (let i = 0; i < strings.length; i++) {
        let layer = stringsParamLayer(strings, i)
        for (let j in paramParsers) {
            let parser = paramParsers[j]
            let lastMatchIndex = -1
            let matches = layer.map((e: string, i: number) => {
                let parsed = parser(e)
                if (parsed) {
                    lastMatchIndex = i
                }
                return parsed
            })
            if (lastMatchIndex != -1) {
                params.push(matches[lastMatchIndex])
                i += lastMatchIndex
                break
            }
        }
    }

    return [
        "command",
        params
    ]
} %}

@{%

const stringsParamLayer = (strings: string[], index: number) => {
    const sliced = strings.slice(index)
    return sliced.map((e: string, i: number) => sliced.slice(0, i + 1).join(""))
}

const paramRegexes: {[key: string]: RegExp} = Object.freeze({
    ws:     /^\s+$/,
    null:   /^(?:null|undefined|NaN)$/,
    bool:   /^(?:false|true)$/,
    word:   /^(?:[A-Za-z]\w*)$/,
    // Other regexes
    txt:    /^(?:[^\s;]+)$/
})

const testParamId = (id: string, text: string) => paramRegexes[id].test(text)

const paramTesters: {[key: string]: Function} = Object.freeze({
    whitespace: (text: string) => testParamId("ws", text),
    null:       (text: string) => testParamId("null", text),
    bool:       (text: string) => testParamId("bool", text),
    word:       (text: string) => testParamId("word", text),
    // Other testers
    text:       (text: string) => testParamId("txt", text)
})

const testAndParseParam = (tester: string, parseFunc: Function, text: string) => paramTesters[tester](text) ? parseFunc(text) : null

// Parse and QuickParse
const parseParamFunc = (type: string) => (text: string) => parse(type, text)
const qarseParamFunc = (type: string) => (text: string) => [type, text]

const paramParsers: {[key: string]: Function} = Object.freeze({
    whitespace: (text: string) => paramTesters.whitespace(text) ? ["whitespace", text]      : null,
    null:       (text: string) => paramTesters.null(text)       ? parse("nullers", text)    : null,
    bool:       (text: string) => paramTesters.bool(text)       ? parse("boolean", text)    : null,
    word:       (text: string) => paramTesters.word(text)       ? ["word", text]            : null,
    // Other parsers
    text:       (text: string) => paramTesters.text(text)       ? ["text", text]            : null
})

%}
```

**Thanks for reading this pull request**

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -239,7 +239,7 @@
             keepHistory: false,
             // Call clone to remove the reference
             // But not at StreamLexer because StreamLexer is defined right here
-            grammar.lexer ? grammar.lexer.clone() : new StreamLexer,
+            lexer: grammar.lexer ? grammar.lexer.clone() : new StreamLexer,
         };
         for (var key in (options || {})) {
             this.options[key] = options[key];
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -234,14 +234,19 @@
         }
         this.grammar = grammar;
 
+        if (!options) {
+            options = {}
+        }
+
         // Read options
         this.options = {
             keepHistory: false,
-            // Call clone to remove the reference
-            // But not at StreamLexer because StreamLexer is defined right here
-            lexer: grammar.lexer ? grammar.lexer.clone() : new StreamLexer,
+            lexer: options.synchronous ?
+                (grammar.lexer ? grammar.lexer.clone() : new StreamLexer) :
+                (grammar.lexer || new StreamLexer),
         };
-        for (var key in (options || {})) {
+        
+        for (var key in options)) {
             this.options[key] = options[key];
         }
 
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -245,8 +245,8 @@
                 (grammar.lexer ? grammar.lexer.clone() : new StreamLexer) :
                 (grammar.lexer || new StreamLexer),
         };
-        
-        for (var key in options)) {
+
+        for (var key in options) {
             this.options[key] = options[key];
         }
 
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -241,9 +241,7 @@
         // Read options
         this.options = {
             keepHistory: false,
-            lexer: options.synchronous ?
-                (grammar.lexer ? grammar.lexer.clone() : new StreamLexer) :
-                (grammar.lexer || new StreamLexer),
+            lexer: grammar.lexer ? (options.synchronous ? grammar.lexer.clone() : grammar.lexer) : new StreamLexer,
         };
 
         for (var key in options) {
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -237,7 +237,9 @@
         // Read options
         this.options = {
             keepHistory: false,
-            lexer: grammar.lexer || new StreamLexer,
+            // Call clone to remove the reference
+            // But not at StreamLexer because StreamLexer is defined right here
+            grammar.lexer ? grammar.lexer.clone() : new StreamLexer,
         };
         for (var key in (options || {})) {
             this.options[key] = options[key];
--- a/docs/index.html
+++ b/docs/index.html
@@ -402,7 +402,9 @@ typechecking primitives for flow-based systems,
 <a href="https://github.com/sizigi/lp5562">lp5562</a> is an assembler for the TI LP5562
 LED driver, <a href="https://github.com/vsl-lang/VSL">VSL</a> is a Versatile Scripting
 Language, <a href="https://github.com/juanlaube/while-typescript">while-typescript</a> is
-an implementation of the WHILE language.</p>
+an implementation of the WHILE language,
+<a href="https://github.com/lo-language/velo">lo</a> is a language for secure distributed
+systems.</p>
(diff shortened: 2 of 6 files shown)


lib/nearley.js
```diff
@@ -234,12 +234,19 @@
...
+            options = {}
+        }
+
         // Read options
         this.options = {
             keepHistory: false,
-            lexer: grammar.lexer || new StreamLexer,
+            lexer: options.synchronous ?
+                (grammar.lexer ? grammar.lexer.clone() : new StreamLexer) :
+                (grammar.lexer || new StreamLexer),
```
akonsu: I propose to change the statement to avoid repetition:
```javascript
...
lexer: grammar.lexer ? (options.synchronous ? grammar.lexer.clone() : grammar.lexer) : new StreamLexer,
...
```

warnio: That's a good idea, I'll change it now in the pull request

warnio: Done

PR #355, opened 2018-03-03
Use Moo lexer to power nearleyc

tjvr:
This PR updates the bootstrapped Nearley grammar to use a Moo lexer. 🐮 

The tricky part was disambiguating charclasses such as `[a-z0-9]` from macro arguments such as `foo[X, Y]`. This is done using a stateful lexer, which transitions to a separate state immediately after a `word`, and back into the `main` state afterwards.

As part of this upgrade, I added some sanity tests which are run on `nearley-language-bootstrapped.ne` immediately before bootstrapping, to check that the parse trees output by the new grammar are the same as the old. These can be found in `test/pre-bootstrap.js`. They're not run as part of the standard test suite, since this would involve comparing the bootstrapped, generated grammar with itself :-)

However, I did include the old scannerless bootstrapped Nearley grammar as a new unit test. The bootstrapped grammar makes use of nearly all our features. The test simply checks that it compiles without throwing any errors.

Although it's not the main rationale for this change, I'll be interested to see if this change makes `nearleyc` any faster. I'm going to throw together some benchmarks for the parsing and compilation stages, separately.


tjvr:
@kach Are you happy for me to merge this? :)

kach:
Yes! Go for it.

PR #309, opened 2017-10-06
Add Travis checks

tjvr:
This adds Travis checks that, for every commit and pull request:

* Checks the tests pass
* Checks the benchmarks don’t throw errors
* Builds the docs
* Checks the working state is clean (ie. the docs are fresh!)

Travis is quick to setup, and having a pass/fail status on every PR gives an extra little reassurance to us maintainers. Please could you set this up @Hardmath123? You just need to log in to <http://travis-ci.org> and enable Travis for hardmath123/nearley. :-)

(The build currently fails, so we can check Travis is working properly! I’ll fix that once this is merged :-))

PR #357, opened 2018-03-04
Remove TypeScript warning

tjvr:
@kach You added a lint warning in c82a2b9 to direct TypeScript users to #287.

This broke the unit tests:
```
  1) bin/nearleyc builds for TypeScript:
     Error: Expected 'WARN\tThe interface for the TypeScript preprocessor changed briefly in nearley 2.10 and 2.11. See https://github.com/Hardmath123/nearley/pull/287 if you encounter issues importing your grammar into TS.\n' to be ''
      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toBe (node_modules/expect/lib/Expectation.js:66:28)
      at Context.<anonymous> (test/nearleyc.test.js:65:24)
```

We either need to amend the unit test, or remove the lint warning. Since some time has passed since the warning as added, this PR removes it.

---

*Aside:* I thought we'd enabled Travis CI, which would warn us about this sort of thing; but [it looks like](https://travis-ci.org/kach/nearley) we didn't. @kach Only you as an admin can enable it; if you wouldn't mind taking 30s to do so, that'd be grand; then I'll merge #309, to turn it on. 😃

### DIFF
--- a/lib/lint.js
+++ b/lib/lint.js
@@ -21,19 +21,9 @@ function lintNames(grm, opts) {
     });
 }
 
-function lintTSVersion(grm, opts) {
-    if (grm.config.preprocessor === 'typescript' &&
-        semver.gt(opts.version, '2.11.0'))
-        warn(opts, "The interface for the TypeScript preprocessor changed " +
-                   "briefly in nearley 2.10 and 2.11. See " +
-                   "https://github.com/Hardmath123/nearley/pull/287 if you " +
-                   "encounter issues importing your grammar into TS.");
-}
-
 function lint(grm, opts) {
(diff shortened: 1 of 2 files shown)


kach:
Yeah, I think it would be fine to remove the warning.

I clicked the button on Travis that says "activate this repository" or whatever. Is that it? And is there a way I can give _you_ admin privileges too?

tjvr:
> I clicked the button on Travis that says "activate this repository" or whatever. Is that it?

Yup, that's it! 😄 https://travis-ci.org/kach/nearley/builds

PR #360, opened 2018-03-06
Disallow single-quoted strings

tjvr:
I misread the original grammar, and wrote a tokenizer which accepts single-quoted strings as well as double-quoted ones.


kach:
👍

Issue #358, opened 2018-03-06
Running out of memory on a small input?


robsimmons:
I don't know what could be going on with this seemingly innocent grammar:

```
@{%

const mooLexer = require('moo').compile({
    space: { match:/[ \t\n]+/, lineBreaks: true },
    ident: {
        match: /[a-z][a-zA-Z0-9_]*/,
        keywords: { keyword: ['fn', 'tfn', 'match', 'ifz', 'forall', 'exists'] }
    },
    sig: 'SIG',
});

%}

@lexer mooLexer

Main    -> _ Tp0 _

Tp1     -> %ident
         | "(" _ Tp0 _ ")"

Tp0      -> Tp1

_       -> %space | null
```

but it appears to run out of memory pretty quickly when given the string "a" to parse, rather than correctly parsing "a" as a valid instance of the grammar.

Full example, check out and run "npm i && npm run test": https://github.com/robsimmons/badgram

tjvr:
Hi! Sorry to hear you're having trouble; thanks for providing details :-)

That's really interesting! I agree your grammar looks normal.

I don't have time to debug further right now; but please could you try Nearley 2.11.1? I'm wondering if it was the change we shipped at the weekend causing you issues. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

robsimmons:
Aaaahhh, yes: everything does work with Nearley 2.12.1. If I'd realized it was a recent update I woulda thought to try that. Thanks!

robsimmons:
I've confirmed that 2.11.1 works for my larger case as well, and added a slightly larger but more realistic grammar example as a test case in the PR above. (Presumably the PR will not pass testing, since it demonstrates the regression 2.11.1 -> 2.12.1.)

kach:
Oh, boy, you're going to love this, Tim: your tokenizer-based compiler compiler is wrong!

```
$ nearleyc types.ne -o types-2.12.1.js
$ npm uninstall -g nearley
removed 10 packages in 0.169s
$ npm install -g nearley@2.11.1
…+ nearley@2.11.1
added 9 packages in 1.152s
$ nearleyc types.ne -o types-2.11.1.js
$ diff types-*.js
1c1
< // Generated automatically by nearley, version 2.11.1
---
> // Generated automatically by nearley, version 2.12.1
26c26,27
<     {"name": "Tp4", "symbols": [{"literal":"("}, "_", "Tp", "_", {"literal":")"}]},
---
[...]

PR #361, opened 2018-03-06
Type testcase

robsimmons:
This is a failing (times out / memory gives up) test case demonstrating Issue #358 . I introduced this test case in a branch off of 2.11.1, where it worked, and then pulled it forward to the head, where it failed.


kach:
Can you clarify which input causes it to hang? It seems to work for me using nearley-test…

```
$ nearley-test infinity.js -qi 'a'
[ { type: 'ident',
    value: 'a',
    toString: [Function: tokenToString],
    offset: 0,
    size: 1,
    lineBreaks: 0,
    line: 1,
    col: 1 } ]
$ nearley-test -v
2.12.1
```

robsimmons:
I believe all succeeding tests loop. Certainly the first one, where it is fed `"a"` as input, loops.

robsimmons:
closing this since it's been eight years

PR #367, opened 2018-04-01
Disallow custom matchers when using a lexer

tjvr:
This is a small change to disallow using custom matchers—functions of the form `{test: (token: Any) -> Boolean}`—when using the `@lexer` directive.

From the updated docs:

> ```js
>  const tokenNumber = { test: x => Number.isInteger(x) };
> ```
> [snip]
>
> Note that this is only supported if you *don't* use `@lexer`. When using
> Nearley with Moo, `%foo` instead matches a *token of type `foo`*. Any JS
> variables named `foo` defined in your grammar have no effect.

I don't believe anyone is using custom matchers together with a tokenizer, and I don't think it would be a particularly useful or sensible thing to do.

Disallowing this makes the code cleaner, allows us to pay down a little bit of technical debt, and should make the parser easier to maintain going forward (_watch this space_...). (Plus, we get to delete some code from Moo!)

### DIFF
--- a/docs/md/tokenizers.md
+++ b/docs/md/tokenizers.md
@@ -90,8 +90,6 @@ However, you can use any lexer that conforms to the following interface:
 - `formatError(token)` returns a string with an error message describing a
   parse error at that token (for example, the string might contain the line and
   column where the error was found).
-- `has(name)` returns true if the lexer can emit tokens with that name. This is
-  used to resolve `%`-specifiers in compiled nearley grammars.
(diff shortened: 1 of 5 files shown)


kach:
"I don't believe anyone is using custom matchers together with a tokenizer"

How sure are we of this?

At this point, I am pretty confused about the different types of things you can put in a nearley rule, and I really don't want to complicate that unless needed.

PR #370, opened 2018-04-13
Syntax highlighting for docs

### LINKED ISSUE #351: Syntax highlight Nearley grammars on the website
We need to do one of:

- switch the highlighting engine to the same one that GitHub uses, which supports Nearley;
- write a plugin for `highlight.js` that understands `nearley` syntax; or
- patch our `highlight-js` integration, in `docs/layouts/template.html`, to interpret the `ne` language as something else (`js` perhaps?)

coolreader18:
This resolves #351.

I switched the markdown parser to the Github Flavored Markdown API, so nearley grammar works with it. If you want to change the syntax theme, you can change the `@import` in the CSS to something generated from [here](https://github.com/primer/github-syntax-theme-generator). It doesn't support smartypants, so hopefully that's not too important.

The only other change is that if you wanted to include something else that's interactive, like on the home page, you'd have to implement it through JS in the template layout, as GFM parses out "dangerous tags" like input. See line 58 in template.html for what I did.

### DIFF
--- a/docs/layouts/template.html
+++ b/docs/layouts/template.html
@@ -114,25 +114,5 @@ function scrollEnd() {
 </script>
 
 <script>
-(function(i, s, o, g, r, a, m) {
-  i["GoogleAnalyticsObject"] = r;
-  (i[r] =
-    i[r] ||
-    function() {
-      (i[r].q = i[r].q || []).push(arguments);
-    }),
-    (i[r].l = 1 * new Date());
-  (a = s.createElement(o)), (m = s.getElementsByTagName(o)[0]);
-  a.async = 1;
-  a.src = g;
-  m.parentNode.insertBefore(a, m);
-})(
-  window,
-  document,
-  "script",
-  "https://www.google-analytics.com/analytics.js",
-  "ga"
-);
-ga("create", "UA-46120535-5", "auto");
-ga("send", "pageview");
+(function(i,s,o,g,r,a,m){i["GoogleAnalyticsObject"]=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date;a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,"script","https://www.google-analytics.com/analytics.js","ga");ga("create","UA-46120535-5","auto");ga("send","pageview");
 </script>
--- a/docs/layouts/template.html
+++ b/docs/layouts/template.html
@@ -136,12 +136,12 @@
     </script>
 
     <script>
-      (function (i, s, o, g, r, a, m) {
-      i['GoogleAnalyticsObject'] = r; i[r] = i[r] || function () {
-        (i[r].q = i[r].q || []).push(arguments)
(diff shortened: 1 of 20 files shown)


docs/layouts/template.html
```diff
@@ -42,73 +42,97 @@ <h2>{{title}}</h2>
...
 <script charset="utf-8" src="/lib/nearley.js"></script>
 <script charset="utf-8" src="/examples/js/csscolor.js"></script>
 <script type="text/javascript">
-window.addEventListener('load', function() {
-    document.getElementById('parse-input').addEventListener('keyup', function() {
+window.addEventListener(
+  "load",
+  function() {
+    document.getElementById("user-content-sample-parse").innerHTML =
+      '<input type="text" id="parse-input" placeholder="rgba(0, 255, 0, 0.5)" />\n\
```
coolreader18: This is how I implemented the sample CSS color parser.

docs/make.js
```diff
@@ -17,51 +17,78 @@ const debug = (files, metalsmith, done) => {
...
-    gfm: true,
-  }))
-  .use(headings('h3'))
-  .use(layouts({
-    engine: 'handlebars',
-    default: 'template.html',
-  }))
+  .use(
+    markdown({
+      // accessToken: "github oath token goes here if being rate-limited"
```
coolreader18: You can put a GitHub OAuth token here if you're being rate-limited while building.

coolreader18:
Also, this means that there are less things to load, as highlight.js isn't a dependency, the classes are all in the HTML.

docs/make.js
coolreader18: And yes, I know OAuth is misspelled.

tjvr:
Thanks for working on this!

Is it possible to not re-format `make.js` please? We prefer pull requests to leave the existing code style of files untouched.

If you could explain what your changes do (e.g. the cheerio part) that would be really helpful too!

I'm on holiday this week, but I'll have a proper look at this when I get back. :-)

coolreader18:
Oh yeah, sorry, I have an atom plugin that automatically formats on save. I'll turn that off and change it, it shouldn't be too bad. The cheerio module is to replace the metalsmith-headings module, I copied the source code from there and modified it a bit to work with github flavored markdown, as it changes ### headings into h3s with a link icon like you'd see in a README.md, and doesn't add an id= to the h3s. metalsmith-headings used cheerio, so that's why I put it in. However, writing the docs should all work the same, except that syntax highlighting works for nearley blocks now!

coolreader18:
In the [marked](https://github.com/markedjs/marked/blob/master/docs/USING_ADVANCED.md#asynchronous-highlighting) docs, there is an option that allows passing a function that handles syntax highlighting, so I might look into using that with a package I published, [github-markdown-syntax](https://npm.im/github-markdown-syntax).

coolreader18:
I just tested it online, so once I'm done with all my finals I'll test it in the make.js file and commit it.

coolreader18:
That should fix it! It looks nearley (pun intended) identical to the previous docs on [nearley.js.org](https://nearley.js.org), the only difference being the colors inside of the code blocks.

coolreader18:
I don't know why the builds are failing, something with `package-lock.json`? The docs do build, on my machine and on travis.

PR #398, opened 2018-09-22
shorten rule strings to reduce minified code size

philschatz:
👋 Thanks for a great parser! I switched to it after using [ohm-js](https://www.npmjs.com/package/ohm-js)

Here is [my grammar](https://github.com/philschatz/puzzlescript-cli/blob/3aff641e62a8e818838e3af1376160d3be201b5b/src/parser/grammar.ne) (part of a project to [make puzzle games more embeddable and accessible](https://github.com/philschatz/puzzlescript-cli) to vision-impaired people).

As a result, the minified code drops from 772k to 200k. I tried using numbers directly but that caused parsing problems which I did not debug.

If this suggestion looks promising, I can try updating the tests and make any changes to the Pull Request.

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -75,13 +75,31 @@
         } else if (s.token) {
             return s.token;
         } else {
-            return JSON.stringify(s);
+            return toCommentedShortName(s);
+        }
+    }
+
+    var lookup = {}
+    var counter = 0
+    function shortenName(name) {
+        if (!lookup[name]) {
+            var str = '' + counter++
+            lookup[name] = str
+        }
+        return lookup[name]
+    }
+
+    function toCommentedShortName(name) {
+        if (typeof name === 'string') {
+            return JSON.stringify(shortenName(name)) + "/*" + name + "*/"
+        } else {
+            return JSON.stringify(name)
         }
     }
 
     function serializeRule(rule, builtinPostprocessors) {
         var ret = '{';
-        ret += '"name": ' + JSON.stringify(rule.name);
+        ret += '"name": ' + toCommentedShortName((rule.name));
         ret += ', "symbols": [' + rule.symbols.map(serializeSymbol).join(', ') + ']';
         if (rule.postprocess) {
             if(rule.postprocess.builtin) {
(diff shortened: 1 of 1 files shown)


kach:
Two comments:
1. Can we make this optional, controlled via a command-line option passed to the compiler? Perhaps `--minify-rule-names`?
2. Can we only minify the "generated" names (with the `$` signs), and leave the human-specified names (like `main`) alone? Because when initializing the Parser you'd otherwise need to know the number to specify which rule to parse.

Also, @tjvr any further thoughts?

philschatz:
I was unable to update the docs. I installed `metalsmith` but after running `cd docs; npm run make; cd ..` I received the following error and am not sure how to proceed:

```
$ cd docs; npm run make

> nearley-docs@0.0.1 make /absolute/path/to/nearley/docs
> rm -f *.html && node make.js && cp index.html ../

/absolute/path/to/nearley/docs/make.js:65
    if (err) throw err;
             ^

Error: no files to process. See https://www.npmjs.com/package/metalsmith-layouts#no-files-to-process
```

tjvr:
Thanks for your contribution!

Have you tried comparing the minified code after gzipping? Files are generally compressed with gzip when they're served over HTTP. I think I would prefer to leave the generated names as-is, and rely on gzip to reduce the file size 🙂

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

philschatz:
Thanks! Indeed, the gzipped file size difference is only 500 bytes for my file. I was looking to put the minified file size in my `README.md` but can use the gzipped size instead ([other sites](https://lodash.com) do this).

I did notice at least one CDN that does not gzip their responses. When requesting https://unpkg.com/puzzlescript-cli@2.0/lib/webpack-output.js Firefox sends an `Accept-Encoding: gzip, deflate, br` Request Header but the response is an un-compressed file (no `Content-Encoding: gzip` Response Header).

An example of the slow request can be found by going to https://philschatz.com/404

But I would expect the vast majority of CDNs to compress their responses.

PR #404, opened 2018-10-27
Added docs for save()/restore()

diegoperini:
I hope my additions reflect the suggested use cases for save and restore operations.

### DIFF
--- a/docs/md/parser.md
+++ b/docs/md/parser.md
@@ -148,7 +148,7 @@ If you plan to recover after an invalid token is fed, it is possible to do so vi
 let parserState = null;
 try {
    parserState = parser.save();
-   parser.feed(...)
+   parser.feed(...);
 } catch (e) {
    parser.restore(parserState);
 }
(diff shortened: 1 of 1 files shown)


diegoperini:
`npm install && npm run make` failed on my computer, thus Travis is angry with me I guess.

abitrolly:
@diegoperini did you run them from the `docs/`? What error have you got?

abitrolly:
This `docker` command shall not fail, because it doesn't not depend on the state of local packages.
```
docker run -v "$(pwd)":"/app" -w "/app/docs" -it --rm node bash -c "npm install && npm run make"
```

gajus:
Can this also be used to speed up parser initialization?, e.g.

```ts
import nearley from 'nearley';
import grammar from './grammar';
import type {
  Ast,
} from './types';

const rules = nearley.Grammar.fromCompiled(grammar);
const parser = new nearley.Parser(rules);
const state = parser.save();

export const parse = (query: string): Ast => {
  parser.restore(state);

  const results = parser.feed(query).results;

  if (results.length === 0) {
    throw new Error('Found no parsings.');
  }

  if (results.length > 1) {
    throw new Error('Ambiguous results.');
  }

  return results[0];
};
```

PR #419, opened 2018-12-20
replace nomnom with commander

kach:
@tjvr Could you review this, please?


kach:
While we're cleaning dependencies, is the only reason we have babel to test "building for ES6+"? Why would this ever fail / can we drop the dependency?

Also our Mocha devDependency seems very out-of-date…

honzajavorek:
@kach I appreciate you've put your time and effort in this 🙇 Thank you!

PR #426, opened 2019-02-07
add babel macro support

Zemnmez:


Zemnmez:
having written all this code, I realise now there's a synchronous method for executing a program I could have used in the babel plugin instead...

PR #442, opened 2019-06-14
PoC for reporting JS syntax errors in postprocessers

airportyh:
This is only a PoC to gather suggestions. It attempts to implement https://github.com/kach/nearley/issues/440

What it doesn't do which I'd like for it to do is report the line number of the error. I believe nearley doesn't currently track line numbers though? I believe that functionality is deferred to the tokenizer. Please correct me if I am wrong about that.


airportyh:
Close for now.

PR #453, opened 2019-07-18
User-friendly error reporting feature V1

airportyh:
First version of implementing: https://github.com/kach/nearley/issues/451

New scheme doesn't require `keepHistory` turned on. It uses back references saved on the state objects. Error reporting code is now integrated directly into the Nearley Parser class.


airportyh:
I ran `npm test` and compared the slower tests with this branch and see no noticeable difference. Here is the result of one of the test runs:

Master:

```
bin/nearleyc

    ✓ builds for ES5 (105ms)

    ✓ builds for ES6+ (993ms)

    ✓ builds for CoffeeScript (247ms)

    ✓ builds for TypeScript (1822ms)

    ✓ builds modules in folders (103ms)

    ✓ builds modules with multiple includes of the same file (104ms)

    ✓ warns about undefined symbol (102ms)
```

This branch:
```
  bin/nearleyc

    ✓ builds for ES5 (104ms)

    ✓ builds for ES6+ (1070ms)

    ✓ builds for CoffeeScript (236ms)

    ✓ builds for TypeScript (1843ms)

    ✓ builds modules in folders (101ms)

    ✓ builds modules with multiple includes of the same file (104ms)

    ✓ warns about undefined symbol (100ms)
```

bin/error-report-experiment.js
```diff
@@ -0,0 +1,17 @@
...
+        nearley.Grammar.fromCompiled(grammar),
+        { keepHistory: false }
+    );
+    try {
+        parser.feed("b4d4ss");
+        console.log("results", parser.results);
+    } catch (e) {
+        console.log(e.message);
+    }
+}
```
kach: Can this be a test instead of a random file in `bin`?

lib/nearley.js
```diff
@@ -338,6 +337,73 @@
...
+    */
+    Parser.prototype.buildStateStack = function(state) {
+        if (!state) {
+            return [];
+        }
+        // console.log(`${state}, wantedBy: ${state.wantedBy.join(", ")}`)
+        if (state.dot === 0) {
+            return [
+                state,
+                ...this.buildStateStack(state.wantedBy[0])
```
kach: Can this be de-sugared out of the `...` syntax? We've tried to keep nearley's core in "old-style" JS throughout and this would be a weird deviation.

lib/nearley.js
```diff
@@ -338,6 +337,73 @@
...
+            throw new Error(`Unknown symbol type: ${symbol}`);
+        }
+    };
+
+    /*
+    Builds the "state stack" - which you can think of as the call stack of the
+    recursive-descent parser which the Nearley parse algorithm simulates.
+    The state stack is represented as an array of state objects. This function
+    needs to be given a starting state, and it returns an array of state objects. 
+    The first item of the array will bet the starting state, with each successive
```
kach: type: `will be[t] the`

lib/nearley.js
```diff
@@ -338,6 +337,73 @@
         return this;
     };
 
+
+    Parser.prototype.reportError = function(token) {
+        var lines = [];
+        const tokenDisplay = (token.type ? token.type + " token: " : "") + JSON.stringify(token.value !== undefined ? token.value : token);
+        lines.push(this.lexer.formatError(token, "Syntax error"));
+        lines.push(`Unexpected ${tokenDisplay}. Instead, I was expecting to see one of the following:\n`);
```
kach: Can these be de-sugared out of the backtick-formatted-string syntax? We've tried to keep nearley's core in "old-style" JS throughout and this would be a weird deviation.

kach:
Very nice. Could you please address the comments I left? Once that's done, I think this is ready for a merge. :-)

kach:
By the way, do you have a blog or something? This is really cool stuff and if you write a short piece showing how the algorithm works and some neat examples of its output, I think people in general would love to see it (I'd also want to add it to the "further reading" section of the documentation).

kach:
One last thing — if you have some extra time, can you try and run this benchmark https://github.com/kach/nearley/blob/master/test/profile.js before/after your change? You may need to modify the script a bit to work. The reason is that I wasn't worried about the speed, but rather the memory impact of your changes. This will be a good sanity check to make sure we haven't accidentally made it significantly more memory-hungry (the script profiles memory usages as well as time).

lib/nearley.js
airportyh: Done.

lib/nearley.js
airportyh: Fixed.

lib/nearley.js
airportyh: Done.

bin/error-report-experiment.js
airportyh: Done.

airportyh:
> One last thing — if you have some extra time, can you try and run this benchmark https://github.com/kach/nearley/blob/master/test/profile.js before/after your change? You may need to modify the script a bit to work. The reason is that I wasn't worried about the speed, but rather the memory impact of your changes. This will be a good sanity check to make sure we haven't accidentally made it significantly more memory-hungry (the script profiles memory usages as well as time).

I ran the profiler in both branches for comparison:

Master:
```
Nearley test.
=============
Tests operate on the grammar p -> "(" p ")" | [a-z]
An input of size n is 2n+1 characters long, and of the form (((...a...))).

Running time tests.
-------------------
Each star corresponds to the time taken to parse an input of that size with a recursive grammar.

SCALE
                    0.25s
                                        0.50s
                                                            0.75s
                                                                                1.00s
0
*
   *
   *
   *
     *
10000
    *
   *
        *
     *
    *
20000
       *
    *
       *
        *
      *
30000
         *
       *
          *
        *
            *
40000
              *
           *
               *
              *
            *
Running memory tests.
-------------------
Each star corresponds to the memory taken to parse an input of that size with a recursive grammar.
Occasional outliers may be caused by gc runs. Nearley profiling doesn't explicitly call the gc before each run.

SCALE
                    025MB
                                        050MB
                                                            075MB
                                                                                100MB
0
+
    +
+
  +
+
10000
    +
         +
       +
           +
              +
20000
+
               +
             +
+
                    +
30000
+
                     +
+
   +
+
40000
                           +
+
+
                             +
+
```

This branch:

```
Nearley test.
=============
Tests operate on the grammar p -> "(" p ")" | [a-z]
An input of size n is 2n+1 characters long, and of the form (((...a...))).

Running time tests.
-------------------
Each star corresponds to the time taken to parse an input of that size with a recursive grammar.

SCALE
                    0.25s
                                        0.50s
                                                            0.75s
                                                                                1.00s
0
*
   *
   *
  *
      *
10000
    *
   *
   *
       *
   *
20000
      *
    *
       *
       *
    *
30000
         *
      *
         *
          *
        *
40000
           *
             *
           *
             *
               *
Running memory tests.
-------------------
Each star corresponds to the memory taken to parse an input of that size with a recursive grammar.
Occasional outliers may be caused by gc runs. Nearley profiling doesn't explicitly call the gc before each run.

SCALE
                    025MB
                                        050MB
                                                            075MB
                                                                                100MB
0
+
+
       +
   +
      +
10000
+
        +
           +
+
                  +
20000
+
                 +
                +
+
                      +
30000
+
+
+
                       +
+
40000
  +
+
                              +
+
                              +
```

airportyh:
Sure, I'd love to blog about this. I'll let you know.

kach:
Awesome, thanks! I'll merge this when I get home.

kach:
Err, it appears getSymbolDisplay doesn't work with custom lexers — could you take a look and fix that? Shouldn't be too hard.

airportyh:
> Err, it appears getSymbolDisplay doesn't work with custom lexers — could you take a look and fix that? Shouldn't be too hard.

Fixed. Could you test again?

airportyh:
I found that in some cases, the error output seems a bit much. Here is an example:

Screen Shot 2019-07-19 at 5 31 25 PM
Screen Shot 2019-07-19 at 5 31 54 PM
Screen Shot 2019-07-19 at 5 32 22 PM

Maybe wait on merging this until I figure out something that avoids this voluminous output.

kach:
Anything "intelligent" scares me, I much prefer the first option, though I'm not quite sure what exactly you'd be passing the `-v` flag to.

You can also consider eliding duplicate lines with "…12 more identical like this" or something.

airportyh:
> Anything "intelligent" scares me, I much prefer the first option, though I'm not quite sure what exactly you'd be passing the `-v` flag to.

You are right, this isn't a command line app necessarily.

> You can also consider eliding duplicate lines with "…12 more identical like this" or something.

I like this idea.

airportyh:
Implemented the code to consider all `wantedBy` items, and to detect cycles and prevent infinite recursion (test included). Even-though it consider all wantedBys, it still only displays one state stack to the user even if there are multiple derivations - To reduce the output amount and also, I think it wouldn't help the user that much to see all derivations. Reverted the change to add the second back reference to the states, when I realized that was unneeded. Also collapsed multiple consecutive identical states in the state stack when displaying. @tjvr @kach Please take a look 🙏

kach:
LGTM. @tjvr you want to take a look? Otherwise I'll merge + publish an updated package tonight Pacific time.

kach:
Okay, I looked it over again and it looks great! I'm going to merge it.

kach:
Thanks for your contribution! It's a very cool feature. If you ever do write a blog post about how it works, I would be very much interested. :-)

airportyh:
Nice! I will let you know if and when I write that blog post. But first, I'd like to move on to next steps: colored output and slight improvement in the display of the input code.

Also, to your knowledge, do you know of a JavaScript parser written in Nearley?

kach:
I think there's the beginnings of a JS parser in nearley in `examples/`, but by no means up-to-date or maintained. I doubt there's a full implementation anywhere.

PR #456, opened 2019-07-28
Colored (when available) output for error reports.

airportyh:
Added colored output for error reports. 

* Uses the colors module when running in Node
* Uses a no-op shim and avoids bundling the colors module when running in the browser

Here is what is looks like at the moment, but I am open to suggestions for changing the visual:
Screen Shot 2019-07-27 at 11 53 19 PM

And in the browser:

Screen Shot 2019-07-27 at 11 58 41 PM

Further work planned:

* currently, error reporting doesn't kick in when a custom lexer throws an error, make that work too (may require additional lexer API method(s))
* allow displaying a few lines of input above the offending line and also highlight the offending token (may require changes in moo)


kach:
A couple questions — why is Travis failing? What is the significance of the pink coloring of the last item?

The other thing is that the `colors` shim is kind of scary. Can we think of alternatives?

airportyh:
> A couple questions — why is Travis failing? What is the significance of the pink coloring of the last item?

Oops, I should be able to fix the Travis error.

I meant for the pink/red coloring to mean that's a symbol that's not yet matched. Yellow means in progress, and green means matched. Open to suggestions here.

> The other thing is that the `colors` shim is kind of scary. Can we think of alternatives?

What about it is scary? The dense-ness of the code? Or the fact that I am using very specific knowledge about how Node works and how the bundlers work?

kach:
Yeah, it seems like a hack (though clever, of course) and I don't think colors are enough of a win to merit putting that in the codebase (not to mention adding another dependency). Sorry.

airportyh:
Ah okay. I'll talk about the implementation vs motivation separately. Also, I'll give some precedent for colored output in terminal programs.

### Implementation

Believe me, I don't like the hacky-ness either, and I have considered a few different options. This one seemed like the best out of the ones I have considered:

* Have colors as a dependency even for the browser, even-though it doesn't do anything. This would require no hack to provide a shim, and so there would only be a single `var colors = require('colors')` line at the top. I ruled this out upon finding out that the colors codebase is larger than Nearley itself: 8.7kB vs 5.1kB, seems like a lot to take on for the users.
* Don't use the colors modules, but instead implement the color output in Nearley itself - this is not difficult to do for one platform (unix), but supporting multiple platforms (Windows also) adds more burden to the maintainers.
* Run Nearley through a build process (using Webpack, probably), and configure the bundler to generate a version for Node vs a different version for web. This requires the work of setting up the bundler.
* The current hack: which detects whether the environment is Node or not. If it is Node, requires the colors module, if it is not Node, does not require the colors module, and instead creates a no-op shim that conforms to the API of the colors module. It relies on the fact that the JS bundlers find require calls with string literals to find modules dependencies, and so if the argument isn't a string literal, it does not attempt to bundle.

There might be other options I haven't thought of too.

### Motivation

I am thinking about teaching Nearley, and the use of color coding is a very powerful technique for education. There has been studies that have shown that the use of color help people retain and retrieve what they've learned (https://www.weareteachers.com/color-coding-classroom/).

I made a video for teaching Nearley using the latest version with the error report and am uploading it to YouTube now. I'd be happy to share when it's done.

### Precedent

The use of colored output in terminal programs is pretty commonplace now. Some of them make better use of color than others, but it's pretty widely accepted.

Here are is an example of git:
Screen Shot 2019-07-29 at 2 07 24 PM

yeoman:
Screen Shot 2019-07-29 at 1 33 11 PM

create-react-app:
Screen Shot 2019-07-29 at 1 42 15 PM

kach:
It seems to me that the specific formatting of the output should be handled by the client — for example, if you're in a browser, you might want to format the parse error via HTML+CSS. So I don't think this should be part of nearley's core anyway, and the need for a hack to get it to work makes it even more concerning.

It looks like your specific use-case for teaching a nearley class can be handled by a separate module that you can pass the error to for colorization (we can also publish that module for other node users to use if they wish).

airportyh:
Yeah for HTML output I was planning on having a separate module to do that, because that definitely seems outside of scope.

I can cope with having a separate module that colorizes the error, maybe nearley-in-color? I just dread that because it's not the main module, it will not be as widely used, and therefore Nearley users who could have been able to benefit from having color, won't (I am not only concerned about *my* students, but *all* students of Nearley). But if you feel strongly about this, I will let it go.

kach:
Okay, great, so it's decided then, we'll have a separate module for error message highlighting. I'm closing this PR then.

I wouldn't worry about userbase. People discover and start using tools soon enough; we'll do a good job making sure they know it exists.

PR #461, opened 2019-08-09
Remove ES6 syntax

### LINKED ISSUE #460: Still ES6 code in lib
I noticed that commit https://github.com/kach/nearley/commit/17a29fce14cce3cb55e4c885521995abda0a6df3 removed ES6+ code and I saw two more.
https://github.com/kach/nearley/blob/17a29fce14cce3cb55e4c885521995abda0a6df3/lib/nearley.js#L347
and
https://github.com/kach/nearley/blob/17a29fce14cce3cb55e4c885521995abda0a6df3/lib/nearley.js#L356
This is not an issue for me personally but it might be for users that still have to support older browsers.

Cheers!

mgroenhoff:
Fixes #460

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -344,7 +344,7 @@
         var lastColumn = this.table[lastColumnIndex];
         var expectantStates = lastColumn.states
             .filter(function(state) {
-                const nextSymbol = state.rule.symbols[state.dot];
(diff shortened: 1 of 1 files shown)


kach:
Excellent, thank you!

PR #471, opened 2019-10-03
Remove nuller from documentation

### LINKED ISSUE #97: using nuller and joiner
Seen [here](https://github.com/Hardmath123/nearley/blob/b3296b31864f60b26237fddae522d18cac7cadc0/lib/generate.js#L118-L123), but if I do something like... `_ -> [\s]:* {% nuller %}` it doesn't work - it says nuller is not defined. I see that you spit out `id` into the generated code, but not joiner, nuller, or arrconcat. How do you use them?

slifty:
As discussed in #97, there are a few postprocessors that are used internally (`joiner`, `arrconcat`, `arrpush`, and `nuller`) but are not accessible from within nearley grammar.

`nuller` was referenced in documentation despite being inaccessible.  This removes that reference so that documentation reflects reality.

Resolves #97

### DIFF
--- a/docs/grammar.html
+++ b/docs/grammar.html
@@ -393,12 +393,10 @@ necessary. You can usually use a tokenizer instead.</p>
 </blockquote>
 </li>
 </ul>
-<p>nearley provides two built-in postprocessors for the most common scenarios:</p>
+<p>nearley provides one built-in postprocessor:</p>
 <ul>
 <li><code>id</code> returns the first element of the <code>data</code> array. This is useful to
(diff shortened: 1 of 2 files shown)


slifty:
I should note -- I opened this PR to allow folks to close the issue, but as a relatively new member of the grammar-definition-user-base there may be some value in listing commonly implemented postprocessors in your documentation.

A list like that is out of my expertise, but may be worth creating an issue around alongside this fix if it feels reasonable to include.

kach:
Thanks!

PR #491, opened 2020-01-15
Fix excessive memory usage for error reporting

airportyh:
Fixes this issue: https://github.com/kach/nearley/issues/475.

I apologize for the wait in getting this problem fixed. The excessive memory usage happened because the error reporter built all possible state stack arriving at a current expectant state (state that is expecting a certain terminal token). In some cases, the number of possible states is so large that node.js is spending a lot of time simply allocating memory for the stack and array allocations. The fix was to not build all possible state stacks, but rather simply build one of them - this should have been done in the first place, because the error reporter was using only the first stack anyway. I added a test to the test suite to proof this worked as well, which is based on a grammar I am working with in a toy programming language called "fun".

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -27,6 +27,7 @@
       "version": "0.1.4",
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
+      "optional": true,
       "requires": {
         "kind-of": "^3.0.2",
         "longest": "^1.0.1",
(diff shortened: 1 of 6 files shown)


kach:
Amazing! Thank you so much. No problem at all about the delay. :-)

I am merging this as-is, but just one question: will the test you introduced fail / take unreasonably long to run if we ever accidentally re-introduce such a "bug"?

airportyh:
> I am merging this as-is, but just one question: will the test you introduced fail / take unreasonably long to run if we ever accidentally re-introduce such a "bug"?

Yes it will, in which case, I suppose Travis will timeout and fail?

PR #497, opened 2020-02-07
using-in-frontend.md doc fix: add `argv: []`

mathieucaroff:
this was necessary to get compilation with nearley 2.19.1 to work
-- at least when run in node.

Without it here's what I get:

```txt
./node_modules/nearley/lib/compile.js:37
                        opts.args[0] ? require('path').dirname(opts.args[0]) : process.cwd(),
                                 ^
TypeError: Cannot read property '0' of undefined
```

### DIFF
--- a/docs/md/using-in-frontend.md
+++ b/docs/md/using-in-frontend.md
@@ -27,7 +27,7 @@ function compileGrammar(sourceCode) {
     const grammarAst = grammarParser.results[0]; // TODO check for errors
 
     // Compile the AST into a set of rules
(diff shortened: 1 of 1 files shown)


kach:
Good catch. Could you rebuild the docs? That's why Travis is failing.

abitrolly:
@mathieucaroff you can rebuild the docs with this command.
```
docker run -v "$(pwd)":"/app":Z -w "/app/docs" -it --rm node bash -c "npm install && npm run make"
```

PR #500, opened 2020-02-21
Unused Parameters in whitespace builtin

### LINKED ISSUE #499: Strict Typescript Options cause error output
```
src/parser/grammar.ts(38,68): error TS6133: 'd' is declared but its value is never read.
src/parser/grammar.ts(41,70): error TS6133: 'd' is declared but its value is never read.
```

The workaround of course is to change my typescript config to:

```
{
  "noUnusedParameters": false
}
```

The offending rules are:

```
ParserRules: [
    // ...
    {"name": "_", "symbols": ["_$ebnf$1"], "postprocess": function(d) {return null;}},
    // ... and
    {"name": "__$ebnf$1", "symbols": ["__$ebnf$1", "wschar"], "postprocess": (d) => d[0].concat([d[1]])},
   // ...
]

It would be nice to be able to avoid this problem though. Thoughts?

jtenner:
fixes #499 

I'm not sure if there's any other places a javascript function like `function(d) {return null;}` occurs, but this is a good start.

I also cannot install the packages on my windows machine, so making the docs will be useless.

Any help?

### DIFF
--- a/builtin/whitespace.ne
+++ b/builtin/whitespace.ne
@@ -1,5 +1,5 @@
 # Whitespace: `_` is optional, `__` is mandatory.
-_  -> wschar:* {% function(d) {return null;} %}
-__ -> wschar:+ {% function(d) {return null;} %}
+_  -> wschar:* {% function() {return null;} %}
+__ -> wschar:+ {% function() {return null;} %}
 
 wschar -> [ \t\n\v\f] {% id %}


hassankhan (approved):

jtenner:
Please let me know if I can help. Thanks!

PR #506, opened 2020-04-29
Bugfix: avoid error while reporting errors with recursive grammars

corwin-of-amber:
Currently, if you have recursive rules in your grammar, and you try to parse some input that has a syntax error, `reportError` crashes with `Cannot read property '0' of null`.

It is obviously tricky to report the full trace in this case, but it seems that at least reporting the state where the error occurred is reasonable.

### DIFF
--- a/lib/nearley.js
+++ b/lib/nearley.js
@@ -356,7 +356,7 @@
         // If there is more than one derivation, we only display the first one.
         var stateStacks = expectantStates
             .map(function(state) {
-                return this.buildFirstStateStack(state, []);
+                return this.buildFirstStateStack(state, []) || [state];
             }, this);
         // Display each state that is expecting a terminal symbol next.
         stateStacks.forEach(function(stateStack) {


kach:
Hey, thanks for this! (Hi Shachar nice to see you again! Last time we talked was four years ago, I think!)

Happy to merge, but for posterity could you show a quick example that fails before this PR and succeeds after this PR? Just a comment here would be enough for future reference.

corwin-of-amber:
Hi Karthik! Yes, it's definitely been ages! I haven't used nearley for a while and I'm glad it's still up and running, and improving too.

I wrote a JS script that converts the Flex/Bison from https://gist.github.com/codebrainz/2933703 to a moo`Lexer` + nearley `Grammar` couple. So the example I currently have is a bit involved. I suspect the same would happen with a simple "calculator" grammar, though, right? Since all that is needed is for the rules to be recursive.

corwin-of-amber:
Yep, this grammar:
```
E -> E "+" F
E -> F

F -> F "*" Q
F -> Q

Q -> [0-9]
Q -> "(" E ")"
```

With the input `1+3a` will give you the error.

corwin-of-amber:
BTW you should totally include an example of a calculator with moo + nearley in your examples folder.

kach:
Perfect, that's exactly what was needed. Thanks!

> BTW you should totally include an example of a calculator with moo + nearley in your examples folder.

Yes, there's a long list of "shoulds" related to examples and documentation… maybe I'll get to it over summer vacation? :-)

kach:
Upgrading nearley from npm to 2.19.3 should have the updated code!

corwin-of-amber:
Woah that was fast lol

PR #512, opened 2020-05-26
Ignore typescript errors in generated file

MarkKoester:
The @ts-ignore flag only suppresses errors on the following line. @ts-nocheck should be used to ignore all errors in the file.


MarkKoester:
I'm also wondering if it would be a good idea to add
```js
/* eslint-disable */
```
to generated files. Right now I get tons of eslint errors from the generated code that I have to manually suppress.

PR #517, opened 2020-06-15
Fixed handling of custom matchers written as test functions

airportyh:
This fixes https://github.com/kach/nearley/issues/487.

Also extracted 2 utility functions getSymbolShortDisplay and getSymbolLongDisplay so that the 2 versions of the symbol display functions are close by each other and future authors can see that they are related.


kach:
Thank you for the quick turnaround! Published an update to npm as well, 2.19.4.

PR #531, opened 2020-08-16
add alternative railroad library

Floffah:
this is a suggestion but you can put this where you like


kach:
You should update the docs by updating the markdown source in `md` and then running `npm run make`. That way the markdown and html stay in sync!

kach:
PS do you have an online example of better-nearley-railroad output?

Floffah:
no but i can make one

Floffah:
I have put 2 examples of outputs [here](https://floffah.github.io/better-nearley-railroad/)

kach:
Very nice! Instead of "(lexer)" you may consider prefixing that symbol with "%".

Once you update the PR to change the markdown and recompile, I'll merge it. Also, to attract more users you can be more specific about the additional features you provide (as opposed to simply "more in-depth").

Floffah:
I cant actually get the nearley dependencies to install but i can give it ago :) also i will write a list of added features rn

Floffah:
for some reason this seems to have bumped the docs version from 2.19.4 to 2.19.5

Floffah:
ok ive tried to fix as much as i could this is confusion

kach:
Okay, I separately added it to the docs. Thanks for your contribution!

Floffah:
thank you!

PR #532, opened 2020-08-19
Link to moo-indentation-lexer module

aliclark:
I've created an NPM module for indentation-aware parsing, which I think would be more useful for linking than the current Gist.

Nb. the command `npm install && npm run make` failed to run on my computer, so the docs still need to be rebuilt.

### DIFF
--- a/docs/tokenizers.html
+++ b/docs/tokenizers.html
@@ -351,8 +351,7 @@ used to resolve <code>%</code>-specifiers in compiled nearley grammars.</li>
 </ul>
 <blockquote>
 <p>Note: if you are searching for a lexer that allows indentation-aware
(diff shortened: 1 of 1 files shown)


Floffah:
> Nb. the command npm install && npm run make failed to run on my computer, so the docs still need to be rebuilt.

Make sure you are using node v10 (you should use NVM (node version manager) to manage node version). Nearley only seems to work using node v10
All you need to do is install nvm and then do `nvm install 10.8.0 && nvm use 10.8.0` (or whatever the latest node v1- version is)

kach:
Cool! Done! (Instead of merging I introduced the link myself.)

PR #537, opened 2020-09-16
Fix path to tokenizers readme

pyoor:


kach:
Ah - the docs are meant to be read online in HTML format, and when compiled the link is correct.
https://nearley.js.org/docs/grammar

PR #540, opened 2020-09-19
Allow whitespace between macro brackets

aliclark:
This change would allow whitespace to be used between square brackets for macros, for example:

```
macro1[ expression ] -> expression
macro1[ expression {% id %} ]
```


kach:
Thanks! I've pushed this, v2.19.7. :)

PR #541, opened 2020-09-20
Add counting lineBreaks in js and btstring

aliclark:
`moo` requires `lineBreaks` to be `true` for it to count the number of newlines within a token (presumably to save performance for the majority of the lexing within lines).

This change improves the accuracy of error messages for Nearley grammars - see https://github.com/kach/nearley/issues/539

### DIFF
--- a/lib/nearley-language-bootstrapped.js
+++ b/lib/nearley-language-bootstrapped.js
@@ -23,6 +23,7 @@ var rules = Object.assign({
     js: {
         match: /\{\%(?:[^%]|\%[^}])*\%\}/,
         value: x => x.slice(2, -2),
+        lineBreaks: true,
     },
     word: {match: /[\w\?\+]+/, next: 'afterWord'},
     string: {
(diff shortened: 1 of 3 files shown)


kach:
This seems reasonable. Did you re-run the tests?

kach:
Let me also loop in @airportyh. :)

aliclark:
The tests passed except for *builds for ES6+* on my machine (I guess that's likely to be unrelated...)

airportyh:
@aliclark would you mind adding a test that would fail without the proposed change? You can add to `test/parser.test.js` and model it after a test like this: https://github.com/kach/nearley/blob/master/test/parser.test.js#L232. If you run into any issue, please ask, and I'd be happy to help you out.

aliclark:
Good idea! Thanks for offering help and pointing to an example. It seemed straightforward enough, how about the test in the commit above?

airportyh:
@aliclark looks great!

kach:
(This is on my radar - I'll take a look this weekend! Thanks for your patience!)

kach:
Hi Ali, hi Toby — thank you for your patience. School got out yesterday and this was at the top of my list. Merged, and I'll push a new release to npm shortly.

Thanks again.

Issue #553, opened 2020-11-05
FYI - nearley ported to Deno


oguimbal:
I ported nearley so it is usable in Deno.

You can find it at [https://deno.land/x/nearley](https://deno.land/x/nearley) Let me know if you have any problem with that :)

(And thank you for this great lib !)

kach:
Cool! If you PR an update to the documentation site I'll merge it. :)

oguimbal:
@kach  Done !

PR #554, opened 2020-11-06
Add Deno + pg-mem doc

oguimbal:
Added some doc, as suggested in #553 

Is that what you had in mind ?

I also added a link to [pg-mem](https://github.com/oguimbal/pg-mem) a lib I recently released that uses Nearley to parse SQL (and execute it in-memory), if that's okay :)


kach:
Thanks for doing this. Could you instead put just a sentence with a link under Tooling >> Other tooling? That's where all "third-party" stuff goes.

Adding the link to pg-mem is okay.

oguimbal:
@kach  No problem, done !

kach:
Thanks!

PR #560, opened 2020-11-28
Fix for TypeScript preprocessor token type incompatibility

### LINKED ISSUE #527: Type incompatibility between nearly and moo

bandaloo:
This is a much less invasive fix that resolves #527

### DIFF
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -193,6 +193,11 @@
         output += parser.customTokens.map(function (token) { return "declare var " + token + ": any;\n" }).join("")
         output += parser.body.join('\n');
         output += "\n";
+        output += "interface NearleyToken {";
+        output += "  value: any;\n";
+        output += "  [key: string]: any;\n";
+        output += "};\n";
+        output += "\n";
         output += "interface NearleyLexer {\n";
         output += "  reset: (chunk: string, info: any) => void;\n";
         output += "  next: () => NearleyToken | undefined;\n";
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -56,7 +56,7 @@
         }
 
         options = options || {};
-        tabulated = lines.map(function addIndent(line, i) {
+        var tabulated = lines.map(function addIndent(line, i) {
             var shouldIndent = true;
 
             if(i == 0 && !options.indentFirst) {
@@ -193,7 +193,7 @@
         output += parser.customTokens.map(function (token) { return "declare var " + token + ": any;\n" }).join("")
         output += parser.body.join('\n');
         output += "\n";
-        output += "interface NearleyToken {";
+        output += "interface NearleyToken {\n";
         output += "  value: any;\n";
         output += "  [key: string]: any;\n";
         output += "};\n";
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -193,7 +193,7 @@
         output += parser.customTokens.map(function (token) { return "declare var " + token + ": any;\n" }).join("")
         output += parser.body.join('\n');
         output += "\n";
-        output += "interface NearleyToken {";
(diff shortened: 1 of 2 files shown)


lib/generate.js
```diff
@@ -56,7 +56,7 @@
         }
 
         options = options || {};
-        tabulated = lines.map(function addIndent(line, i) {
+        var tabulated = lines.map(function addIndent(line, i) {
```
kach, line 59: Good catch!

lib/generate.js
```diff
@@ -193,16 +193,11 @@
         output += parser.customTokens.map(function (token) { return "declare var " + token + ": any;\n" }).join("")
         output += parser.body.join('\n');
         output += "\n";
-        output += "interface NearleyToken {";
-        output += "  value: any;\n";
-        output += "  [key: string]: any;\n";
-        output += "};\n";
-        output += "\n";
```
kach, line 200: Wait… are you sure this should be deleted? Isn't `NearleyToken` referenced in the type signature for `next()` a couple lines below?

kach:
Thanks for proactively submitting this PR! See my comments.

lib/generate.js
bandaloo: um yeah not sure what i was thinking sorry. somehow i ran the wrong test

kach:
Looks good! Last request: Is there a test you can _add_ that passes after this change but fails before this change?

bandaloo:
I added a test that only passes with the `never` change to the generator, which I learned is only needed with the `--strict` compilation mode (but I think that's how most people use TypeScript anyways). In doing so, I also realized that the original typescript test does not fail if there are type errors! That's because `tsc` has no problem handing you a js file even if there were type errors, and the type errors get logged to stdout. I fixed this too. (That's how I made the mistake of lopping off the `NearleyToken` interface and felt confident because the tests still passed, because type errors weren't caught in the tests.)

kach:
Merged! Thanks so much!

Let's watch out for any TS-related issues reported in the next few days… I'm always super paranoid about changes like this. But I think we - really, _you_ - did a great job here. :)

PR #567, opened 2021-01-14
Allows javascript keywords as literals

### LINKED ISSUE #525: Parser compilation fails if token names collide with JavaScript keywords
I'm rewriting the grammar for my language to use a lexer (:cow:) and i run in to this massively annoying problem.

So I had this:
```
compound ->
    assignment
  | function_call
  | "return" _ expr                                                           {% ast.makeReturn %}
```

And changed it to: 
```
compound ->
    assignment
  | function_call
  | %return _ expr                                                           {% ast.makeReturn %}
```
Where `%return` is a token from the lexer.

I then got this error:
```
SyntaxError: /home/grandmother/git/Changtopia/changtopia/changlang/compiled_grammar.js: Unexpected token (41:79)

 39 |     {"name": "compound", "symbols": ["assignment"]},
  40 |     {"name": "compound", "symbols": ["function_call"]},
> 41 |     {"name": "compound", "symbols": [(lexer.has("return") ? {type: "return"} : return), "_", "expr"], "postprocess": ast.makeReturn},
```
As you can se it has taken `%return` and just used `return` as the name of the variable in the parser.

The same appears to be happening for any token name that collides with a JavaScript keyword.

Is the only solution to this to just name my tokens differently?

AsLogd:
Looking at the generated typescript type for the tokens I guessed the real intent, so please correct me if I guessed wrong. In my understanding there were missing quotes when building tokens for the parser rules. Fixing this caused the typescript tests to fail, because the (incorrectly) used variables where given `any` type, thus making the ternary expression adopt this broader type and passing the test. Then I fixed the type to what I think it should be.

fixes #525

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -131,7 +131,7 @@
                     if (result.customTokens.indexOf(name) === -1) {
                         result.customTokens.push(name);
                     }
-                    var expr = result.config.lexer + ".has(" + JSON.stringify(name) + ") ? {type: " + JSON.stringify(name) + "} : " + name;
+                    var expr = result.config.lexer + ".has(" + JSON.stringify(name) + ") ? {type: " + JSON.stringify(name) + "} : \"" + name + "\"";
                     return {token: "(" + expr + ")"};
                 }
                 return token;
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -131,7 +131,7 @@
                     if (result.customTokens.indexOf(name) === -1) {
                         result.customTokens.push(name);
                     }
-                    var expr = result.config.lexer + ".has(" + JSON.stringify(name) + ") ? {type: " + JSON.stringify(name) + "} : \"" + name + "\"";
+                    var expr = result.config.lexer + ".has(" + JSON.stringify(name) + ") ? {type: " + JSON.stringify(name) + "} : " + JSON.stringify(name);
                     return {token: "(" + expr + ")"};
                 }
                 return token;
--- a/lib/generate.js
+++ b/lib/generate.js
@@ -190,7 +190,6 @@
         output +=  "// Bypasses TS6133. Allow declared but unused functions.\n";
         output +=  "// @ts-ignore\n";
         output += "function id(d: any[]): any { return d[0]; }\n";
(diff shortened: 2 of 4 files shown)


kach:
Hmm… this seems reasonable. But perhaps you want to use `JSON.stringify` instead of `"\""+…+"\""` concatenation to quote the string, as the rest of that line of code does?

(cc @tjvr)

AsLogd:
We could use template literals. [Support seems to go as far as 4.9](https://node.green/#ES2015-syntax-template-literals-basic-functionality). The line would look like this:  
```js
var expr = `${result.config.lexer}.has("${name}") ? {type: "${name}"} : "${name}"`;
``` 
I could also change other occurrences of JSON.stringify if you like that. Else, I would just use `stringify` to match the current style. It's your call.

AsLogd:
Hi, is there something I could do to speed things up? This fix could be pretty convenient for me right now,

kach:
Thanks for the reminder, and sorry for the delay. (It's hard maintaining a package for free as a full-time student…)

Can you change it to indeed use JSON.stringify as discussed above?

AsLogd:
No problem. I'm truly grateful for your work and I'm impressed that you could develop it while studying, I'm not super familiar with nearley, but feel free to tag me on other issues/pull requests if you need another set of eyes.

I made the requested change, Just to confirm, the 'nuller' test is not passing on master, right? (is not passing on this branch neither)

kach:
The `nuller` test is skipped — it isn't run at all.

AsLogd:
Is this ready to be merged then?

kach:
Ah, so I took another look the other and actually I think the current behavior is correct and indeed what is intended: if the lexer doesn't has() a token with that name, then the next fallback is to check if there is a variable with that name in scope that describes the token. I think this was a backwards-compatibility-related thing.

The right move is probably to emit a warning (in the linter) if you try to use a keyword as the name of a lexer token… someday I'll get around to doing that… :)

AsLogd:
I see. All right. Thanks

4the1appsllc:
# _**```
```

```
```**_

PR #578, opened 2021-04-16
TypeScript type definitions

corwin-of-amber:
(still working on this branch, just letting you know)

I imported the definitions from DefinitelyTyped and added the missing constructor.
I am working on porting their tests to use `tsd`, theirs use `dtslint` which is a huge overkill.

### DIFF
--- a/lib/nearley.d.ts
+++ b/lib/nearley.d.ts
@@ -1,0 +1,120 @@
+// Type definitions for runtime classes
+//  - Parser (and ParserOptions)
+//  - Lexer (and Token, LexerState)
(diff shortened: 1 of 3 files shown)


corwin-of-amber:
BTW I see `coffee-script` referenced in `coffee-script` but no `.coffee` files. Looks like `docopt` is written in CoffeeScript but is already compiled to JS in the `npm` package, so perhaps this dep can be removed?

corwin-of-amber:
I added the test files and they look ok, to run them all that needs to be done is `npx tsd`. Should I add this command to the `npm run test` script?

Issue #400, opened 2018-09-22
. does not match '\u2028'


raine:
Not sure if intentional. If so, how can I match it?

```ne
start -> .
```

```js
const nearley = require('nearley')
const grammar = require('./test')

const parser = new nearley.Parser(
  nearley.Grammar.fromCompiled(grammar)
)

parser.feed('\u2028')
```

```
Error: invalid syntax at line 1 col 1:

   
  ^
Unexpected " "
```

tjvr:
I'm not sure, but perhaps you want `[^]` rather than `.`?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

raine:
Interesting. How does that work and why isn't it same as `.`?

tjvr:
It's a quirk of JS RegExps. Quoting from the readme of a different project: 

> the dot /./ doesn't include newlines. Use [^] instead if you want to match newlines too.
> 
> Since an excluding character ranges like /[^ ]/ (which matches anything but a space) will include newlines, you have to be careful not to include them by accident!

The character class `[^]` matches anything. `^` in a character class means "anything but", so following it with an empty set just means "anything".

The dot `.` excludes newlines by default, unless you use the dotAll modifier, which Nearley doesn't. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

abitrolly:
@kach is there a place in docs to document dot `.` behavior and list things like `[^]`? I came to ask the same question - why `.` doesn't match newlines.

Issue #348, opened 2018-02-11
Match any string


scratchyone:
I'm trying to make a grammar. I wanted to have something in quotes, and I figured that I would match any string inside of quotes, but I can't figure out how to do it.
My current code for detecting a string is:
`char -> [.]
string -> char:*`
I need it to work with any valid piece of JavaScript code. Thanks

scratchyone:
Just fixed it. Instead of a dot for regex, I had to use [^]

PR #587, opened 2021-08-26
Add a small syntax cheatsheet

abitrolly:
I grouped all Nearley matching constructs in one block to make them easier to find and compare with other languages. 

  * `"if"` - match `if` literally
  * `"if"i` - match `if` case-insensitive
  * `.` - match any symbol except newline
  * `[^]` - match anything including newline
  * `[a-zA-Z]` - match symbols in range
  * `[^a-zA-Z]` - match symbols not in range
  * `:+` - match previous construction one or more time
  * `:*` - match zero times or more
  * `:?` - zero time or one time

As I am editing this from the GitHub, I can not preview and format it to proper HTML, but I hope it makes documentation already a bit more useful. The completely undocumented thing was `[^]` and `.` behavior from #348 and #400.

### DIFF
--- a/docs/grammar.html
+++ b/docs/grammar.html
@@ -324,15 +324,18 @@ following nonterminal matches zero or more <code>cow</code>s in a row, such as
 <p>Syntax for matching input if you’re not using a tokenizer.</p>
 <ul>
 <li><code>&quot;if&quot;</code> - match <code>if</code> literally</li>
-<li><code>&quot;if&quot;i</code> - match <code>if</code> case-insensitive</li>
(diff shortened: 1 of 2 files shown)


abitrolly:
I've found that there are `.md` files that I should edit instead of `.html`. Fixed now.

As a side note, have you considered telling GitHub to serve content from the `docs/` folder?

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

That would make it a little bit more intuitive for new users..

abitrolly:
Preview https://abitrolly.github.io/nearley/docs/grammar.html#matching-rules

PR #611, opened 2022-04-13
Allow the user to pick an output language

DavidTanner:
I would like to be able to use the cli to generate typescript output.  This change adds an optional language flag to specify which of the existing generators might be used.


DavidTanner:
Awesome.  Thank you for the answer.  That took a bit of digging to find in the docs.  https://nearley.js.org/docs/parser#using-preprocessors

PR #642, opened 2023-08-21
unparse.js: don't clear rule depth cache for massive speedup

seansfkelley:
I noticed that the unparse script was brutally slow for even modestly-sized grammars and modest depths. I would typically see 5-10s generation times for a grammar with 30-40 rules and depth 14.

`min_depths_rule` appears to be a memoizing cache, and also seems to be cleared _very_ frequently. I don't think it ever needs to be cleared, and by avoiding doing so, my generation times drop to a few milliseconds (!). I toyed around a bit and anecdotally the results seem comparable.

However, I don't actually understand the unparsing algorithm; my claim for the correctness of this change rests purely on manual code analysis. To wit: the values in this cache should never change, no matter how deeply in the tree the function is called, as there are only 3 inputs to the depth calculation: `i` (once for each rule), `visited` (always starts as `[]`) and the cache itself.


seansfkelley:
I'd be happy to restructure this code to avoid the kinda-fishy dependency of the function on `min_depth_rule_cache` from the outer scope, if you'd like, but for the time being I opted to keep the diff small and simple instead of undertaking a larger rewrite.

